PR #22, opened 2018-06-06
Split immutable and persistent interfaces

### LINKED ISSUE #1: Rename from immutable to persistent
There is some immediate backlash about immutable collections by Java users, as their understanding of the concept is either getting a full copy per operation, or receiving an exception on mutating methods.

The `persistent` keyword along with some references to Big O perf may help spread the understanding of the difference.

ilya-g:
This PR splits immutable collection interfaces each into two parts: immutable interface which just declares collection immutability, and persistent one which adds modification operations and builder on top of that.

We faced several times when it would be advantageous to declare that a collection is immutable, however the need to implement all the modification operations made us desist from that.

For example `ImmutableList.subList` definitely returns an immutable list, however it could be not an easy task to implement all the modifications and builder in that sublist. Same for `ImmutableMap.keys/values/entries` collections.

If merged this will fix #1

### DIFF
--- a/kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/AbstractImmutableMap.kt
+++ b/kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/AbstractImmutableMap.kt
@@ -19,7 +19,7 @@ package kotlinx.collections.immutable
 import org.pcollections.PMap
 import java.util.ConcurrentModificationException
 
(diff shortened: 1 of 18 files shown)


armarti:
@ilya-g I'm wondering if you expect this to be merged in the near future? I'm wanting to use this library in a project, and would prefer to use this branch if you expect it will be the way the library is written going forward.

PR #33, opened 2019-03-31
Support fail-fast behaviour in iterators and add tests from guava

qurbonzoda:


kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/persistentOrderedMap/PersistentOrderedMapBuilder.kt
```diff
@@ -64,44 +64,43 @@ internal class PersistentOrderedMapBuilder<K, V>(private var map: PersistentOrde
...
             val nextLinks = mapBuilder[links.next]!!
-            assert(nextLinks.previous == key)
-            mapBuilder[links.next] = LinkedValue(nextLinks.value, links.previous, nextLinks.next)
+//            assert(nextLinks.previous == key)
+            mapBuilder[links.next as K] = nextLinks.withPrevious(links.previous)
         }
-        firstKey = if (key == firstKey) links.next else firstKey
-        lastKey = if (key == lastKey) links.previous else lastKey
+        firstKey = if (links.previous === EndOfLink) links.next as? K else firstKey
+        lastKey = if (links.next === EndOfLink) links.previous as? K else lastKey
```
ilya-g: Makes sense to move these to corresponding `else` blocks, to avoid `firstKey = firstKey` assignment.

qurbonzoda: +1

kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/persistentOrderedMap/PersistentOrderedMapBuilder.kt
```diff
@@ -93,14 +93,16 @@ internal class PersistentOrderedMapBuilder<K, V>(private var map: PersistentOrde
             val previousLinks = mapBuilder[links.previous]!!
 //            assert(previousLinks.next == key)
             mapBuilder[links.previous as K] = previousLinks.withNext(links.next)
+        } else {
+            firstKey = links.next as? K
```
ilya-g: Why safe cast here? If `next`/`previous` is not that sentinel object then it's definitely of type `K`. Though the cast is unchecked anyway, replacing it with safe cast isn't a way to get rid of the warning.

qurbonzoda: The if statement above checks only `previous`. `next` may be the sentinel object regardless of `previous`.

qurbonzoda: Suppose map contained a single element, the links `previous` and `next` are both the sentinel object.

ilya-g: Ok, but then this check won't work as expected given that `K` is erased to upper bound (`Object`) at run time

kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/immutableMap/PersistentHashMapBuilder.kt
```diff
@@ -24,7 +24,13 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
     internal var marker = Marker()
     internal var node = map.node
     internal var operationResult: V? = null
+    internal var modCount = 0
+
     override var size = map.size
+        set(value) {
+            field = value
+            modCount++
+        }
```
ilya-g: Is it possible that operations that do not modify size (e.g. `Map.put`) nevertheless invalidate internal structure of the map, so its iterators are becoming invalid after that?

kotlinx-collections-immutable/tests/src/test/kotlin/kotlinx.collections.immutable/contractTests/GuavaImmutableCollectionBaseTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import junit.framework.AssertionFailedError
+import junit.framework.TestListener
+import junit.framework.TestResult
+import junit.framework.TestSuite
+import java.io.PrintWriter
+import java.io.StringWriter
+import kotlin.test.fail
+
+open class GuavaImmutableCollectionBaseTest: TestListener {
+    override fun addFailure(test: junit.framework.Test, e: AssertionFailedError) {
```
ilya-g: Why not just rethrow these exceptions here?

kotlinx-collections-immutable/build.gradle
```diff
@@ -25,6 +25,7 @@ allprojects {
     dependencies {
         testCompile "org.jetbrains.kotlin:kotlin-test-junit:$kotlin_version"
         testCompile "junit:junit:4.12"
+        testCompile 'com.google.guava:guava-testlib:18.0'
```
ilya-g: Should be added to the dependencies of `tests` project only

kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/persistentOrderedMap/PersistentOrderedMapBuilder.kt
qurbonzoda: Now that the `firstKey` and `lastKey` are assumed to hold either instance of `K` or `LinkEnd`, this safe cast is unnecessary.

kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/immutableMap/PersistentHashMapBuilder.kt
qurbonzoda: It turns out it's possible. Only in case of `Map.put` and `List.set`, so I fixed them.

kotlinx-collections-immutable/build.gradle
qurbonzoda: +1

qurbonzoda:
Indentation fixed

kotlinx-collections-immutable/tests/src/test/kotlin/kotlinx.collections.immutable/contractTests/GuavaImmutableCollectionBaseTest.kt
qurbonzoda: +1

PR #37, opened 2019-04-25
Rename Marker to MutabilityOwnership

ilya-g:
The idea is to rename `Marker` class, `marker` parameter and unify these classes of persistent List, Set and Map.

### DIFF
--- a/kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/immutableMap/TrieNode.kt
+++ b/kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/immutableMap/TrieNode.kt
@@ -16,6 +16,8 @@
 
 package kotlinx.collections.immutable.implementations.immutableMap
 
+import kotlinx.collections.immutable.internal.MutabilityOwnership
+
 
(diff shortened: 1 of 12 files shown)


qurbonzoda:
Is unifying Marker classes going to be in a separate PR?

ilya-g:
If there are no objections or better ideas on the naming, I'll continue the same renaming in the other implementations in this PR.

kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable/implementations/immutableMap/TrieNode.kt
```diff
@@ -311,11 +311,11 @@ internal class TrieNode<K, V>(
         return TrieNode(dataMap, nodeMap xor positionMask, newBuffer)
     }
 
-    private fun mutableRemoveNodeAtIndex(nodeIndex: Int, positionMask: Int, mutatorMarker: Marker): TrieNode<K, V>? {
+    private fun mutableRemoveNodeAtIndex(nodeIndex: Int, positionMask: Int, mutatorMarker: MutabilityOwnership): TrieNode<K, V>? {
```
qurbonzoda: `mutatorMarker` -> `owner` as in other places

PR #40, opened 2019-06-21
Benchmark other immutable collections frameworks

qurbonzoda:


vlsi:
Other JVM libraries: com.github.andrewoma.dexx:dexx-collections:0.2, org.pcollections:pcollections:3.0.3, com.github.krukow:clj-ds:0.0.4.

DmitryNekrasov:
The comparative harness is worth having for the 1.0 performance work, but this branch is too far from the current build to revive: 116 files against the pre-migration layout. Opened #269 to redo it fresh, with the extra libraries suggested here noted there. Closing in favor of that.

PR #48, opened 2019-07-22
Comment benchmark methods

qurbonzoda:


qurbonzoda:
Ready

DmitryNekrasov:
These comments target the benchmarks-mpp sources, which have since been restructured, so the diff no longer applies, and the comments alone aren't worth redoing against the new layout. Closing.

Issue #45, opened 2019-07-03
Investigate iteration slowdown in case of collisions in persistent sets/maps


qurbonzoda:
Canonicalization of collision nodes may help.

DmitryNekrasov:
PR #52 tried separating collision nodes into a polymorphic CollisionTrieNode, but that regressed hashSet performance and was dropped (now closed). ilya-g's suggestion there is the direction to try next: cut the asymptotic overhead by checking isCollision only in leaf nodes on the operation path, instead of at every node.

PR #52, opened 2019-08-08
Set and map collisions optimization

qurbonzoda:
Benchmark parameters adjustment and benchmark result files will be removed before merging.

Introducing `CollisionTrieNode` for collision nodes (polymorphism approach) adds significant performance losses, see hashSet regressions: https://teamcity.jetbrains.com/viewLog.html?buildId=2469110&buildTypeId=KotlinTools_KotlinxCollectionsImmutable_BenchmarkHashIterations&tab=artifacts#%2FbenchmarkResults
Thus, collision nodes and normal nodes were not separated into different entities.

Final benchmark results: https://teamcity.jetbrains.com/viewLog.html?buildId=2469130&buildTypeId=KotlinTools_KotlinxCollectionsImmutable_BenchmarkHashIterations&tab=artifacts


qurbonzoda:
Also addresses https://github.com/Kotlin/kotlinx.collections.immutable/issues/45.

ilya-g:
Let's try to eliminate asymptotic overhead on big sets/maps. For example, given the current representation of a collision node, I believe `isCollision` could be checked only in leaf nodes on the operation path.

DmitryNekrasov:
Closing: the polymorphic CollisionTrieNode approach here regressed hashSet performance, and the branch is conflicting against the current build. Kept the problem and the next direction to try in #45.

Issue #68, opened 2020-01-21
ConcurrentModificationException when using replaceAll() on PersistentMap.builder()


Virtlink:
Calling `replaceAll()` on the mutable builder of a `PersistentMap` throws a `ConcurrentModificationException`.

## Reproduction
To reproduce:

	val pmap = persistentMapOf("a" to 1, "b" to 2)
	val mmap = pmap.builder()
	mmap.replaceAll { _, u -> u + 1 }	// throws ConcurrentModificationException
	mmap.build()

## What I expected

A persistent map with the mapping `"a" to 2, "b" to 3`.

## What I got

An exception:

```
java.util.ConcurrentModificationException
	at kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilderLinksIterator.checkForComodification(PersistentOrderedMapBuilderContentIterators.kt:57)
	at kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilderLinksIterator.next(PersistentOrderedMapBuilderContentIterators.kt:26)
	at kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilderEntriesIterator.next(PersistentOrderedMapBuilderContentIterators.kt:69)
	at kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilderEntriesIterator.next(PersistentOrderedMapBuilderContentIterators.kt:61)
	at java.util.Map.replaceAll(Map.java:675)
```

qurbonzoda:
This issue was fixed in 0.3.2 version of the library.
Thank you for reporting this bug!

Virtlink:
This issue is not fixed. The above code still throws a `ConcurrentModificationException` in version 0.3.4.

PR #73, opened 2020-03-27
Fix map iterators

qurbonzoda:
Also addresses https://github.com/Kotlin/kotlinx.collections.immutable/issues/68

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
@@ -8,32 +8,37 @@ package kotlinx.collections.immutable.implementations.immutableMap
 import kotlinx.collections.immutable.internal.assert
 
 
-internal class TrieNodeMutableEntriesIterator<K, V>(private val builder: PersistentHashMapBuilder<K, V>)
-    : TrieNodeBaseIterator<K, V, MutableMap.MutableEntry<K, V>>() {
+internal class TrieNodeMutableEntriesIterator<K, V>(
+        private val parentIterator: PersistentHashMapBuilderEntriesIterator<K, V>
+) : TrieNodeBaseIterator<K, V, MutableMap.MutableEntry<K, V>>() {
 
     override fun next(): MutableMap.MutableEntry<K, V> {
         assert(hasNextKey())
         index += 2
         @Suppress("UNCHECKED_CAST")
-        return MutableMapEntry(builder, buffer[index - 2] as K, buffer[index - 1] as V)
+        return MutableMapEntry(parentIterator, buffer[index - 2] as K, buffer[index - 1] as V)
     }
 }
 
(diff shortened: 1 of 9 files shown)


core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
```diff
@@ -8,32 +8,37 @@ package kotlinx.collections.immutable.implementations.immutableMap
...
+private class MutableMapEntry<K, V>(
+        private val parentIterator: PersistentHashMapBuilderEntriesIterator<K, V>,
+        key: K,
+        override var value: V
+) : MapEntry<K, V>(key, value), MutableMap.MutableEntry<K, V> {
+
     override fun setValue(newValue: V): V {
         val result = value
         value = newValue
-        builder[key] = newValue
```
qurbonzoda: Here is the bug https://github.com/Kotlin/kotlinx.collections.immutable/issues/68.
The value is updated bypassing the iterator this mutable entry was obtained from.

core/commonMain/src/implementations/immutableMap/PersistentHashMapContentIterators.kt
```diff
@@ -144,6 +147,7 @@ internal abstract class PersistentHashMapBaseIterator<K, V, T>(node: TrieNode<K,
             if (i > 0) {
                 path[i - 1].moveToNextNode()
             }
+            path[i].reset(TrieNode.EMPTY.buffer, 0)
```
qurbonzoda: Here we release iterated nodes to avoid memory leaks

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -194,28 +201,14 @@ internal class TrieNode<K, V>(
     }
 
     /** The given [newNode] must not be a part of any persistent map instance. */
-    private fun mutableUpdateNodeAtIndex(nodeIndex: Int, positionMask: Int, newNode: TrieNode<K, V>, owner: MutabilityOwnership): TrieNode<K, V> {
+    private fun mutableUpdateNodeAtIndex(nodeIndex: Int, newNode: TrieNode<K, V>, owner: MutabilityOwnership): TrieNode<K, V> {
 //        assert(buffer[nodeIndex] !== newNode)
 
-        val newNodeBuffer = newNode.buffer
-        if (newNodeBuffer.size == 2 && newNode.nodeMap == 0) {
```
qurbonzoda: Here canonicalisation of trie can change order of iteration (see below), therefore I removed canonicalisation for mutable maps as the update may be made by an iterator currently iterating the map builder.

-------------------------               iterator.remove() after                           --------------------------
| k1 | v1 | k3 | v3 | child1 |           iterator.next() returned (k4, v4)               | k1 | v1 | k2 | v2 | k3 | v3 |
--------------------|----                                                                                --------------------------
                                  |                                     -->
             ------------------
             | k4 | v4 | k2 | v2 |
             ------------------

Issue #76, opened 2020-06-13
PersistentOrderedSetIterator breaks if hash of an element changes


Loomie:
# Problem
```
Exception in thread "main" kotlin.KotlinNullPointerException
	at kotlinx.collections.immutable.implementations.persistentOrderedSet.PersistentOrderedSetIterator.next(PersistentOrderedSetIterator.kt:22)
```
When iterating over a `PersistentOrderedSet`

# How to reproduce
After creating the PersistentSet alter a contained element so its hash code changes.
```
import kotlinx.collections.immutable.PersistentSet
import kotlinx.collections.immutable.toPersistentSet

fun main() {
    val changing = mutableSetOf("ok")
    val persistent: PersistentSet<Any> = setOf("constant", changing, "fix").toPersistentSet()
    doSomething(changing, persistent)
    changing.add("break iteration")
    doSomething(changing, persistent)
}

private fun doSomething(changing: MutableSet<String>, persistent: PersistentSet<Any>) {
    println("actual hash is ${changing.hashCode()}")
    val result = persistent.firstOrNull { it is Set<*> }
    println("Found $result")
}
```
Version used: 0.3.2

# Details
The error occurs in `TrieNode.get(Int, K, Int)` because although the data is in the buffer `hasEntryAt()` returns `false`. Seems like the `dataMap` is outdated.

Note: The set is not changed during iteration but during two independent iterations.

I found no documentation on PersistentSet that values must not change their hash codes. Like a Set it should always be iterable.

ilya-g:
We believe, it's generally not a good idea to put data with changing hash code to a hash based immutable data structure. Therefore we're going to detect such situation during set iteration and throw `ConcurrentModificationException` when it happens.

PR #81, opened 2020-08-17
PersistentOrderedSetIterator breaks if hash of an element changes #76

qurbonzoda:


core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
```diff
@@ -28,7 +28,10 @@ internal open class PersistentOrderedMapBuilderLinksIterator<K, V>(
         lastIteratedKey = nextKey
         nextWasInvoked = true
         index++
-        val result = builder.hashMapBuilder[nextKey]!!
+        @Suppress("UNCHECKED_CAST")
+        val result = builder.hashMapBuilder.getOrElse(nextKey as K) {
```
ilya-g, line 32: Observation: the unchecked cast is the result of an inconsistency between `get` and `getOrElse` functions for Map. Perhaps, it's worth addressing that in stdlib.

ilya-g (approved):

core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
```diff
@@ -28,7 +28,10 @@ internal open class PersistentOrderedMapBuilderLinksIterator<K, V>(
...
+            throw ConcurrentModificationException("Hash code of a key ($nextKey) has changed after it was added to the persistent map.")
```
ilya-g, line 33: Will we get the same message for a persistent set backed by this map? Better to make it more generic then, e.g. "... to the persistent collection."

ilya-g: Oh, I see the changes for set below

Issue #83, opened 2020-08-18
Faster union/intersection implementation for sets?


belyaev-mikhail:
Current union (`addAll`) implementation for sets is O(logN * M) as far as I understand even when both sides have the same implementation.
It is certainly possible to improve that to log or even sub-log implementation for HAMTs by merging one-bit arrays only.
It is also possible to improve intersection, although atm you don't even have a method for intersection in the API.

Not sure if it helps maps in any way, but I guess it does.
While intersection is questionable, union is a very important and widely used operation (especially so when `+` is overloaded as union) for persistent sets and implementing this doesn't even need changing the api.

qurbonzoda:
It would be great.

belyaev-mikhail:
So, #86 is the first version of the pull request.
I've done some benchmarking (not the stuff I do often, so critique is welcome) and these are preliminary results:

https://docs.google.com/spreadsheets/d/1NaMgasBhqyThazU-_TVWAK2fCcnztAQ9WCRxr9lNfUw/edit?usp=sharing
https://docs.google.com/spreadsheets/d/1lTjcPPfOZTKd4eHHrzrLOoq9SbaNf9-eEx4MCsyzfMU/edit?usp=sharing

These look promising, but there currently are a few pitfalls:

- Current implementation doesn't try its extreme to preserve persistense (namely, it does not return the same node if all the children are reference-equal to the previous node), because it's much trickier to do for `addAll`/`putAll` efficiently;
- I still struggle to design a solution for ordered maps/sets, because `addAll` cannot be implemented using `HashMap` `putAll` due to different conflict behaviour (`putAll` replaces older values, while `addAll` should retain older values, but edit the new ones);
- The whole `DeltaCounter` business may seem sloppy and less Kotlin'y than your `ModificationResult` approach, but I found massive boxing for bulk operations too heavy and it's not trivial to implement bulk operations this way (sometimes you modify against the _other_ node, not `this`);
- No `mutableAddAll`/`mutablePutAll` for now because I'm not sure if they provide any increased benefit (and, tbh, I still don't fully get how to get around your `MutabilityOwnership` correctly);
- No intersection yet, because that would involve changing the current interfaces.

Feel free to check out and destroy the PR =)

belyaev-mikhail:
Some comments on the benchmarks: I tested two scenarios, _typical_ and _worst-case_ where typical is uniting two non-intersecting sets/maps and worst case is two completely identical, but different, sets.
[...]

PR #86, opened 2020-08-27
Faster union operations

belyaev-mikhail:
Pull request for #83

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
@@ -5,7 +5,10 @@
 
 package kotlinx.collections.immutable.implementations.immutableMap
 
-import kotlinx.collections.immutable.*
+import kotlinx.collections.immutable.ImmutableCollection
+import kotlinx.collections.immutable.ImmutableSet
+import kotlinx.collections.immutable.PersistentMap
+import kotlinx.collections.immutable.mutate
 
 internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
                                        override val size: Int): AbstractMap<K, V>(), PersistentMap<K, V> {
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
@@ -6,6 +6,7 @@
 package kotlinx.collections.immutable.implementations.immutableMap
 
 import kotlinx.collections.immutable.PersistentMap
+import kotlinx.collections.immutable.internal.DeltaCounter
 import kotlinx.collections.immutable.internal.MutabilityOwnership
 
 internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap<K, V>) : PersistentMap.Builder<K, V>, AbstractMutableMap<K, V>() {
@@ -59,6 +60,18 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
         operationResult = null
         node = node.mutablePut(key.hashCode(), key, value, 0, this)
         return operationResult
+    }
+
+    override fun putAll(from: Map<out K, V>) {
+        val map = from as? PersistentHashMap ?: (from as? PersistentHashMapBuilder)?.map
+        if (map != null) @Suppress("UNCHECKED_CAST") {
+            val intersectionCounter = DeltaCounter()
+            val oldSize = this.size
+            node = node.mutablePutAll(map.node as TrieNode<K, V>, 0, intersectionCounter, this)
+            val newSize = oldSize + map.size - intersectionCounter.count
+            if(oldSize != newSize) this.size = newSize
+        }
+        else super.putAll(from)
     }
 
     override fun remove(key: K): V? {
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -5,7 +5,10 @@
 
 package kotlinx.collections.immutable.implementations.immutableMap
 
+import kotlinx.collections.immutable.internal.DeltaCounter
 import kotlinx.collections.immutable.internal.MutabilityOwnership
+import kotlinx.collections.immutable.internal.assert
+import kotlinx.collections.immutable.internal.forEachOneBit
 
 
 internal const val MAX_BRANCHING_FACTOR = 32
@@ -436,6 +439,87 @@ internal class TrieNode<K, V>(
         return this
     }
 
+    private fun mutableCollisionPutAll(otherNode: TrieNode<K, V>,
+                                       intersectionCounter: DeltaCounter,
+                                       owner: MutabilityOwnership): TrieNode<K, V> {
+        assert(nodeMap == 0)
+        assert(dataMap == 0)
+        assert(otherNode.nodeMap == 0)
+        assert(otherNode.dataMap == 0)
+        val tempBuffer = this.buffer.copyOf(newSize = this.buffer.size + otherNode.buffer.size)
+        var i = this.buffer.size
+        for (j in 0 until otherNode.buffer.size step ENTRY_SIZE) {
+            @Suppress("UNCHECKED_CAST")
+            if (!this.collisionContainsKey(otherNode.buffer[j] as K)) {
+                tempBuffer[i] = otherNode.buffer[j]
+                tempBuffer[i + 1] = otherNode.buffer[j + 1]
+                i += ENTRY_SIZE
+            } else intersectionCounter.count++
+        }
+
+        return when (val newSize = i) {
+            this.buffer.size -> this
+            otherNode.buffer.size -> otherNode
+            tempBuffer.size -> TrieNode(0, 0, tempBuffer, owner)
+            else -> TrieNode(0, 0, tempBuffer.copyOf(newSize), owner)
+        }
+    }
+
+    private fun mutablePutAllFromOtherNodeCell(other: TrieNode<K, V>,
+                                               positionMask: Int,
+                                               shift: Int,
+                                               intersectionCounter: DeltaCounter,
+                                               mutator: PersistentHashMapBuilder<K, V>): TrieNode<K, V> {
+        return when {
+            other.hasNodeAt(positionMask) -> {
+                mutablePutAll(
+                        other.nodeAtIndex(other.nodeIndex(positionMask)),
+                        shift + LOG_MAX_BRANCHING_FACTOR,
+                        intersectionCounter,
+                        mutator
+                )
+            }
+            other.hasEntryAt(positionMask) -> {
+                val keyIndex = other.entryKeyIndex(positionMask)
+                val key = other.keyAtIndex(keyIndex)
+                val value = other.valueAtKeyIndex(keyIndex)
+                val oldSize = mutator.size
+                val newNode = mutablePut(
+                        key.hashCode(),
+                        key,
+                        value,
+                        shift + LOG_MAX_BRANCHING_FACTOR,
+                        mutator
+                )
+                if (mutator.size == oldSize) {
+                    intersectionCounter.count++
+                }
+                newNode
+            }
+            else -> this
+        }
+    }
+
+    private fun calculateSize(): Int {
+        if (nodeMap == 0) return buffer.size / ENTRY_SIZE
+        val numValues = dataMap.countOneBits()
+        var result = numValues
+        for(i in (numValues * ENTRY_SIZE) until buffer.size) {
+            result += nodeAtIndex(i).calculateSize()
+        }
+        return result
+    }
+
+    private fun elementsIdentityEquals(otherNode: TrieNode<K, V>): Boolean {
+        if (this === otherNode) return true
+        if (nodeMap != otherNode.nodeMap) return false
+        if (dataMap != otherNode.dataMap) return false
+        for (i in 0 until buffer.size) {
+            if(buffer[i] !== otherNode.buffer[i]) return false
+        }
+        return true
+    }
+
     fun containsKey(keyHash: Int, key: K, shift: Int): Boolean {
         val keyPositionMask = 1 shl indexSegment(keyHash, shift)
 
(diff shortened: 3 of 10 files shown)


core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
```diff
@@ -63,6 +64,14 @@ internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
     }
 
     override fun putAll(m: Map<out K, @UnsafeVariance V>): PersistentMap<K, V> {
+        if(m is PersistentHashMap) @Suppress("UNCHECKED_CAST") {
```
ilya-g, line 67: minor, style: space between `if` and `(`, here and below

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -277,6 +304,94 @@ internal class TrieNode<E>(
...
+                // no element on left -> pick right
+                hasNoCellAt(positionMask) -> otherNode.buffer[otherNodeIndex]
+                // no element on right -> pick left
+                otherNode.hasNoCellAt(positionMask) -> buffer[thisIndex]
+                // both nodes contain something at the masked bit
+                else -> {
+                    val thisCell = buffer[thisIndex]
+                    val otherNodeCell = otherNode.buffer[otherNodeIndex]
+                    val thisIsNode = thisCell is TrieNode<*>
+                    val otherIsNode = otherNodeCell is TrieNode<*>
```
ilya-g, line 337: Does extracting these `is` checks into variables brings benefits here? 
We have to repeat the cast again in the when branches below (I suppose due to this smart-cast not being supported by Kotlin DFA).

belyaev-mikhail: Well, this way we do it two times (check here and cast later), but otherwise we (potentially) do it up to four times (if we inline these variables to the cases in `when`, they are checked in sequence, producing three `is` checks). AFAIK, kotlin compiler doesn't try to do any optimizations for repeating logic in when cases.

Actually, the smart casts do work, but they cast to `TrieNode<*>` instead of `TrieNode<E>`. I also am not sure that having a smart cast avoids having a hard cast in backend code (I'm pretty sure in JVM it inserts a check anyways).

core/commonTest/src/contract/set/ImmutableSetTest.kt
```diff
@@ -10,13 +10,50 @@ import tests.contract.compare
...
     override fun <T> immutableSetOf(vararg elements: T) = persistentHashSetOf(*elements)
     override fun <T> testBuilderToPersistentSet(builder: PersistentSet.Builder<T>) {
         assertNotSame(builder.build(), builder.toPersistentSet(), "toPersistent shouldn't call build()")
     }
+
+    @Test fun addAllElements() {
+        run {
+            val left = immutableSetOf<Int>() + (1..2000)
+            compareSets(left, left.addAll(immutableSetOf()))
+            compareSets(left, immutableSetOf<Int>().addAll(left))
```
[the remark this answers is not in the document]
belyaev-mikhail: Good point

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -258,6 +261,52 @@ internal class TrieNode<E>(
         return this
     }
 
+    private fun collisionAddAll(otherNode: TrieNode<E>, intersectionSizeRef: DeltaCounter): TrieNode<E> {
```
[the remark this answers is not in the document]
belyaev-mikhail: I've tried doing exactly this just yesterday in my more experimental branch and running experiments now: https://github.com/belyaev-mikhail/kotlinx.collections.immutable/blob/faster-intersection-operations/core/commonMain/src/implementations/immutableSet/TrieNode.kt#L285

It was nearly char-to-char equal to your code before I started abstracting things)

core/commonTest/src/contract/set/ImmutableSetTest.kt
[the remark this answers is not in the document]
qurbonzoda: Though the second one is not required to be reference equal. That is `empty + pSet` is not required to be `=== pSet`.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
ilya-g: Another option is to do `as?` instead of `is` and compare the result with null in when branches, e.g.
```
val thisAsNode = thisCell as TrieNode<E>
...
when {
    thisAsNode != null -> thisAsNode.doSomethingWithNode()
```

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -277,6 +326,194 @@ internal class TrieNode<E>(
...
+        if (shift > MAX_SHIFT) {
+            return mutableCollisionAddAll(otherNode, mutator)
+        }
+        // union mask contains all the bits from input masks
+        val newBitMap = bitmap or otherNode.bitmap
+        // first allocate the node and then fill it in
+        // we are doing a union, so all the array elements are guaranteed to exist
+        val mutableNode = when {
+            newBitMap == bitmap && ownedBy == mutator.ownership -> this
+            newBitMap == otherNode.bitmap && otherNode.ownedBy == mutator.ownership -> otherNode
```
qurbonzoda, line 435: This case never occurs. Namely, `otherNode.ownedBy == mutator.ownership` to be true `otherNode` has to be a node in `mutator.node`, because ownership objects are unique to each builder. 
Thus, `otherNode.ownedBy == mutator.ownership` means `builder.addAll(builder)` has been called. So, `this === otherNode` and the method returns without reaching this point.
Anyway, modifying argument is conceptually not correct for `addAll` operation.

core/commonTest/src/contract/set/ImmutableSetTest.kt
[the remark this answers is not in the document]
belyaev-mikhail: > That is empty + pSet is not required to be === pSet.

It is not required (or true) with current impl, it will be true with mine afaik, but conceptually I think it should not be required.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
belyaev-mikhail: Well, for me (conceptually) a union operation is more like `+`, being symmetrical and all, rather than having a receiver and an argument, so I try to treat both arguments equally.
While it makes sense for `+`, it does not make much of it for the builder variant, in that you are correct.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -277,6 +326,194 @@ internal class TrieNode<E>(
...
+        // union mask contains all the bits from input masks
+        val newBitMap = bitmap or otherNode.bitmap
+        // first allocate the node and then fill it in
+        // we are doing a union, so all the array elements are guaranteed to exist
+        val newNode = TrieNode<E>(newBitMap, arrayOfNulls<Any?>(newBitMap.countOneBits()))
+        // for each bit set in the resulting mask,
+        // either left, right or both masks contain the same bit
+        // Note: we shouldn't overrun MAX_SHIFT because both sides are correct TrieNodes, right?
+        newBitMap.forEachOneBit { positionMask ->
+            val newNodeIndex = newNode.indexOfCellAt(positionMask)
```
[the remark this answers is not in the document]
belyaev-mikhail: Or maybe just always have an index to `forEachOneBit`, it makes the code a little bit cleaner and, afair, there is not much use for it without index.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -277,6 +360,192 @@ internal class TrieNode<E>(
...
+                                    mutator.ownership
+                            )
+                        }
+                    }
+                }
+            }
+        }
+        return when {
+            this.elementsEquals(mutableNode) -> this
+            otherNode.elementsEquals(mutableNode) -> otherNode
```
qurbonzoda, line 544: While for small objects `equals` method works fast, for bigger objects, like collections, these equality checks could be tangible. Considering https://github.com/Kotlin/kotlinx.collections.immutable/pull/86#discussion_r485081529 and #Comment we can suppose changes by comparing original `mutator.size` and the size we get after merging the two nodes.
Proof-of-concept commit: https://github.com/Kotlin/kotlinx.collections.immutable/commit/a07b56335903f9fc9f827f8f60b125fbbd719f56

core/commonMain/src/implementations/immutableSet/PersistentHashSetBuilder.kt
```diff
@@ -43,6 +43,16 @@ internal class PersistentHashSetBuilder<E>(private var set: PersistentHashSet<E>
         return size != this.size
     }
 
+    override fun addAll(elements: Collection<E>): Boolean {
+        if(elements is PersistentHashSet) {
```
qurbonzoda, line 47: The optimized union algorithm could also be used for two builders.

core/commonMain/src/implementations/immutableSet/PersistentHashSet.kt
```diff
@@ -21,6 +22,13 @@ internal class PersistentHashSet<E>(internal val node: TrieNode<E>,
...
 
     override fun addAll(elements: Collection<E>): PersistentSet<E> {
+        if (elements is PersistentHashSet) {
+            val intersectionCountRef = DeltaCounter()
+            val newNode = node.addAll(elements.node, 0, intersectionCountRef)
+            if (node === newNode) return this
+            if (elements.node === newNode) return elements
+            return PersistentHashSet(newNode, size + elements.size - intersectionCountRef.count)
+        }
         return this.mutate { it.addAll(elements) }
```
qurbonzoda, line 32: Does optimizing `addAll` for `PersistentHashSet` significantly improves performance now that `builder.addAll` is optimized?
Looking through the implementations they seem almost identical. If the performance impact is not significant, optimizing only for builder is more favourable.

qurbonzoda: Sacrificing a throwaway builder to avoid duplication of big non-trivial code seems like a bargain.

core/commonMain/src/implementations/immutableSet/PersistentHashSetBuilder.kt
```diff
@@ -43,6 +43,16 @@ internal class PersistentHashSetBuilder<E>(private var set: PersistentHashSet<E>
...
+            val oldSize = this.size
+            size += elements.size
```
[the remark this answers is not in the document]
belyaev-mikhail: I don't like it, too, actually, it is unnatural even when you don't consider `modCount`.
But calling `calculateSize()` is supposed to be expensive (it's linear in the size of the node) and my variant only calls it in marginal cases, while yours does it a lot.
It needs to be benchmarked.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
belyaev-mikhail: `elementsEquals()` (bad naming) does not call `equals()` at all. It only checks the reference-equality of the elements, so it's not important what the elements really are.

core/commonMain/src/implementations/immutableSet/PersistentHashSet.kt
belyaev-mikhail: Fair point. If we can work out the current issues of builder implementation, having only one method is a good goal to match.

belyaev-mikhail: Especially considering that `retainAll`, `removeAll` and `containsAll` may also be optimized (the last one having huge speedup as per my experiments) and having two versions of each of these methods would be hell.

core/commonMain/src/implementations/immutableSet/PersistentHashSetBuilder.kt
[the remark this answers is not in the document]
belyaev-mikhail: I've actually put a lot of thought into it precisely with the goal of avoiding calling `calculateSize()` unnecessarily as it seems kinda big deal to me and the `size--` scheme, while not good by any means, is the only thing I came up with.
It is certainly possible to devise an ideal solution here (see the `DeltaCounter` approach in the non-mutable variant) that somehow takes sizes of the roots into account, but it needs more thought.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
belyaev-mikhail: Maybe call it `elementsIdentityEquals`?)

belyaev-mikhail: And yes, it produces virtually no overhead, I benchmarked it quite thoroughly.

core/commonMain/src/implementations/immutableSet/PersistentHashSetBuilder.kt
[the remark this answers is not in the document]
belyaev-mikhail: May be a decent tradeoff would be to carry both an intersection counter and the mutator, overwriting the actual size of the mutator in the end. Just a thought.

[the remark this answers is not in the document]
qurbonzoda: To reproduce `ConcurrentModificationException`:
```
builder.addAll(persistentHashSetOf(1))
iterator.next()  <---
```
No elements were added and internal structure also haven't been changed.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
qurbonzoda: Oh, I see.
In that case the second check, `otherNode.elementsEquals(mutableNode)`, makes less sense. See the elements equality check:
```
// both are just E => compare them
thisCell == otherNodeCell -> thisCell
```
In all cases element of `this` node is selected meaning the check is true only if all cells in `this` node are contained in `otherNode` by reference, which is very rare.

The first check, `this.elementsEquals(mutableNode)`, returns true if all elements in `otherNode` are contained in `this` node by `==`. This makes more sense to me. Therefore, it would be great if we find another way to check this condition (Hopefully after https://github.com/Kotlin/kotlinx.collections.immutable/pull/86#discussion_r485980991 gets resolved).

My main concern is the complexity this code and `elementsEquals` method introduce. It seems to me unjustified.

core/commonMain/src/implementations/immutableSet/PersistentHashSetBuilder.kt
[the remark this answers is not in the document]
belyaev-mikhail: Yes, but this is a very questionable example. One could argue that CME should be thrown in this case because the code is wrong by structure. It is very fragile and should not be written like this.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
belyaev-mikhail: > In all cases element of this node is selected meaning the check is true only if all cells in this node are contained in otherNode by reference, which is very rare.

```kotlin
persistenSetOf(1,2,3) + persistentSetOf(1,2,3,4)
```

Is it rare? How?

> My main concern is the complexity, and maybe redundancy, this code and elementsEquals method introduce. 

I thought about it a lot and the only other option I see to preserve persistence is to add additional checks inside the loop, which is worse because it complicates an already complicated logic. This method has a clear purpose and doing it post-factum has a clean intent.

belyaev-mikhail: Another argument is if we actually try to optimize other bulk operations (which get more benefit from merging than `addAll`), this method is universally suitable for all of them, but implicit logic that preserves persistence in other ways is not.

belyaev-mikhail:
- Implemented mutableAddAll using hybrid DeltaCounter/Builder approach
- Removed addAll and implemented it using mutableAddAll (benchmarks show it works fine)

core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
```diff
@@ -63,6 +64,14 @@ internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
     }
 
     override fun putAll(m: Map<out K, @UnsafeVariance V>): PersistentMap<K, V> {
+        if (m is PersistentHashMap) @Suppress("UNCHECKED_CAST") {
```
qurbonzoda, line 67: Let's optimize `PersistentHashMapBuilder.putAll` and use it here as it was done in `PersistentHashSet.addAll `.

belyaev-mikhail:
Done the same with putAll

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -436,6 +439,81 @@ internal class TrieNode<K, V>(
...
+        val numValues = dataMap.countOneBits()
+        var result = numValues
+        for(i in (numValues * ENTRY_SIZE) until buffer.size) {
+            result += nodeAtIndex(i).calculateSize()
+        }
+        return result
+    }
+
+    private fun elementsEquals(otherNode: TrieNode<K, V>): Boolean {
+        if (this === otherNode) return true
```
qurbonzoda, line 514: `newNode` is a newly created object and cannot be `===` to `this` node or `otherNode`.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -436,6 +439,81 @@ internal class TrieNode<K, V>(
...
+        if (this.buffer === otherNode.buffer) return true
```
qurbonzoda, line 515: this check also seems redundant

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -436,6 +439,81 @@ internal class TrieNode<K, V>(
...
+        assert(nodeMap == 0)
+        assert(dataMap == 0)
+        assert(otherNode.nodeMap == 0)
+        assert(otherNode.dataMap == 0)
+        var res = this
+        for(i in 0 until otherNode.buffer.size step ENTRY_SIZE) {
+            val k = otherNode.keyAtIndex(i)
+            val v = otherNode.valueAtKeyIndex(i)
+
+            val mres = res.collisionPut(k, v)
```
qurbonzoda: Let's optimize this method as `collisionAddAll` was optimized.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -258,6 +279,54 @@ internal class TrieNode<E>(
...
+                is TrieNode<*> -> e.calculateSize()
+                else -> 1
+            }
+        }
+        return result
+    }
+
+    private fun elementsIdentityEquals(otherNode: TrieNode<E>): Boolean {
+        if (this === otherNode) return true
+        if (this.buffer === otherNode.buffer) return true
```
qurbonzoda, line 322: This check can't return true

core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
```diff
@@ -61,6 +62,18 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
...
+    override fun putAll(from: Map<out K, V>) {
+        val map = from as? PersistentHashMap ?: (from as? PersistentHashMapBuilder)?.map
+        if (map != null) @Suppress("UNCHECKED_CAST") {
+            val intersectionCounter = DeltaCounter()
+            val oldSize = this.size
+            node = node.mutablePutAll(map.node as TrieNode<K, V>, 0, intersectionCounter, this)
+            val newSize = oldSize + map.size - intersectionCounter.count
+            if(oldSize != newSize) this.size = newSize
+        }
+        super.putAll(from)
```
qurbonzoda, line 74: This should be in `else` block or the `if` above should return

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -436,6 +439,88 @@ internal class TrieNode<K, V>(
...
+        assert(nodeMap == 0)
+        assert(dataMap == 0)
+        assert(otherNode.nodeMap == 0)
+        assert(otherNode.dataMap == 0)
+        val tempBuffer = this.buffer.copyOf(newSize = this.buffer.size + otherNode.buffer.size)
+        var j = 0
+        for (i in this.buffer.size until tempBuffer.size step ENTRY_SIZE) {
+            @Suppress("UNCHECKED_CAST")
+            if (!this.collisionContainsKey(otherNode.buffer[j] as K)) {
+                tempBuffer[i] = otherNode.buffer[j]
```
qurbonzoda, line 454: Seems like the indices got mixed up.
This should probably have looked like `tempBuffer[j] = otherNode.buffer[i - this.buffer.size]`.
Another option is to iterate through `otherNode.buffer`:
```
         var j = 0
         val otherBuffer = otherNode.buffer
         for (i in 0 until otherBuffer.size step ENTRY_SIZE) {
             @Suppress("UNCHECKED_CAST")
             if (!this.collisionContainsKey(otherBuffer[i] as K)) {
                 tempBuffer[j] = otherBuffer[i]
                 tempBuffer[j + 1] = otherBuffer[i + 1]
                 j += ENTRY_SIZE
             }
             ...
```

core/commonMain/src/implementations/immutableMap/TrieNode.kt
qurbonzoda: This comment was relevant in immutable implementation

core/commonMain/src/implementations/immutableMap/TrieNode.kt
qurbonzoda: As we never create `mutableNode` with buffer of another node

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -477,6 +562,103 @@ internal class TrieNode<K, V>(
...
+                else -> {
+                    val oldKeyIndex = otherNode.entryKeyIndex(positionMask)
+                    mutableNode.buffer[newKeyIndex] = otherNode.keyAtIndex(oldKeyIndex)
+                    mutableNode.buffer[newKeyIndex + 1] = otherNode.valueAtKeyIndex(oldKeyIndex)
+                    if (this.hasEntryAt(positionMask)) intersectionCounter.count++
+                }
+            }
+        }
+        return when {
+            this.elementsEquals(mutableNode) -> this
```
qurbonzoda, line 656: Let's rename the method to `elementsIdentityEquals` to have consistent naming with set `TrieNode`

core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
```diff
@@ -6,6 +6,7 @@
 package kotlinx.collections.immutable.implementations.immutableMap
 
 import kotlinx.collections.immutable.*
+import kotlinx.collections.immutable.internal.DeltaCounter
```
qurbonzoda, line 9: Not needed any more

core/commonMain/src/implementations/immutableSet/TrieNode.kt
qurbonzoda: This one wasn't fixed

qurbonzoda (approved):
Thank you for the great effort! Incredible job!
Let's squash all these commits into 1 or 2 commits and merge to the master.

belyaev-mikhail:
Hope I didn't mess something up during squash

core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
```diff
@@ -61,6 +62,18 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
         return operationResult
     }
 
+    override fun putAll(from: Map<out K, V>) {
+        val map = from as? PersistentHashMap ?: (from as? PersistentHashMapBuilder)?.map
```
ilya-g, line 66: Hmm, this looks suspicious: `(from as? PersistentHashMapBuilder)?.map`. This returns the original immutable map from which builder was created. When we later get the root `node` of that map, it may be different from the root node of the builder if it has been already modified.
Please add a test for that.

belyaev-mikhail:
This seems like a bug to me: we should use the node and not the map. Same goes for set builder, I guess.

belyaev-mikhail:
Actually, no. It's not the right way to do it either. Using the node property will allow modifying the result immutable map undetectably through the right-hand-side builder. I am not sure what's the right solution here.
An assertion? Calling `build()` instead?

 @qurbonzoda

belyaev-mikhail:
#89

PR #90, opened 2020-10-19
Faster intersection operations

belyaev-mikhail:
Ok, this is a more major change than the previous one, but possibly more profitable. Also note that this changes public interface of `ImmutableSet`, adding `retainAll`/`intersect` pair that our domain is particularly interested in.

In addition to union, intersection, difference and containment bulk ops are all implemented using the same strategies.

`containsAll` is particularly desirable, because it has a significant speedup (up to 50x)

Benchmark results:
https://docs.google.com/spreadsheets/d/1NaMgasBhqyThazU-_TVWAK2fCcnztAQ9WCRxr9lNfUw/edit?usp=sharing

### DIFF
--- a/core/commonMain/src/ImmutableCollection.kt
+++ b/core/commonMain/src/ImmutableCollection.kt
@@ -68,6 +68,16 @@ public interface PersistentCollection<out E> : ImmutableCollection<E> {
     fun removeAll(predicate: (E) -> Boolean): PersistentCollection<E>
 
     /**
+     * Returns all elements in this collection that are also
+     * contained in the specified [elements] collection.
+     *
+     * @return a new persistent set with elements in this set that are also
+     * contained in the specified [elements] collection;
+     * or this instance if no modifications were made in the result of this operation.
+     */
+    fun retainAll(elements: Collection<@UnsafeVariance E>): PersistentCollection<E>
+
+    /**
      * Returns an empty persistent collection.
      */
     fun clear(): PersistentCollection<E>
--- a/core/commonMain/src/extensions.kt
+++ b/core/commonMain/src/extensions.kt
@@ -291,6 +291,26 @@ operator fun <E> PersistentSet<E>.minus(elements: Array<out E>): PersistentSet<E
 operator fun <E> PersistentSet<E>.minus(elements: Sequence<E>): PersistentSet<E>
         = mutate { it.removeAll(elements) }
 
+/**
+ * Returns all elements in this set that are also
+ * contained in the specified [elements] collection.
+ *
+ * @return a new persistent set with elements in this set that are also
+ * contained in the specified [elements] collection;
+ * or this instance if no modifications were made in the result of this operation.
+ */
+infix fun <E> PersistentSet<E>.intersect(elements: Iterable<E>): PersistentSet<E>
+        = if (elements is Collection) retainAll(elements) else mutate { it.retainAll(elements) }
+
+/**
+ * Returns all elements in this collection that are also
+ * contained in the specified [elements] collection.
+ *
+ * @return a new persistent set with elements in this collection that are also
+ * contained in the specified [elements] collection
+ */
+infix fun <E> PersistentCollection<E>.intersect(elements: Iterable<E>): PersistentSet<E>
+        = this.toPersistentSet().intersect(elements)
 
 /**
  * Returns the result of adding an entry to this map from the specified key-value [pair].
--- a/core/commonMain/src/implementations/immutableSet/PersistentHashSet.kt
+++ b/core/commonMain/src/implementations/immutableSet/PersistentHashSet.kt
@@ -6,7 +6,6 @@
 package kotlinx.collections.immutable.implementations.immutableSet
 
 import kotlinx.collections.immutable.PersistentSet
-import kotlinx.collections.immutable.internal.DeltaCounter
 import kotlinx.collections.immutable.mutate
 
 internal class PersistentHashSet<E>(internal val node: TrieNode<E>,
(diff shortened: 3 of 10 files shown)


core/commonMain/src/ImmutableCollection.kt
```diff
@@ -67,6 +67,16 @@ public interface PersistentCollection<out E> : ImmutableCollection<E> {
      */
     fun removeAll(predicate: (E) -> Boolean): PersistentCollection<E>
 
+    /**
+     * Returns all elements in this set that are also
```
qurbonzoda, line 71: in this ~set~ **collection**

core/commonMain/src/extensions.kt
```diff
@@ -291,6 +291,27 @@ operator fun <E> PersistentSet<E>.minus(elements: Array<out E>): PersistentSet<E
...
+
+/**
+ * Returns all elements in this collection that are also
+ * contained in the specified [elements] collection.
+ *
+ * @return a new persistent collection with elements in this set that are also
+ * contained in the specified [elements] collection;
+ * or this instance if no modifications were made in the result of this operation.
+ */
+infix fun <E> PersistentCollection<E>.intersect(elements: Iterable<E>): PersistentCollection<E>
```
qurbonzoda, line 313: Kotlin intersect function has the following signature:
```
infix fun <T> Iterable<T>.intersect(other: Iterable<T>): Set<T>
```
That is the function always returns a `Set`.
Hence returning `PersistentSet<E>` here would be more consistent with stdlib.

qurbonzoda (changes requested):
Please add some `containsAll` tests

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -41,13 +41,17 @@ private fun Array<Any?>.removeCellAtIndex(cellIndex: Int): Array<Any?> {
     return newBuffer
 }
 
+// marker value for "empty" results, used in intersection implementation
+private val NOVALUE = Any()
```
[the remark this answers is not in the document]
belyaev-mikhail: That's a possibility, didn't think about it tbh

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -305,6 +309,52 @@ internal class TrieNode<E>(
...
     }
 
+    private fun mutableCollisionRetainAll(otherNode: TrieNode<E>, intersectionSizeRef: DeltaCounter,
+                                          owner: MutabilityOwnership): Any? {
+        if (this === otherNode) {
+            intersectionSizeRef += buffer.size
+            return this
+        }
+        val tempBuffer =
+                if (owner == ownedBy && buffer.size < otherNode.buffer.size) buffer
```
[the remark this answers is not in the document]
belyaev-mikhail: The logic was as follows: Intersection size is the min of two input sizes. So we can reuse the buffer if we know for sure that it's less than the other buffer.

It should be "less or equal" actually.

But thinking of it again, seems like it does not matter, we can try to reuse the buffer in any situation.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -447,6 +497,223 @@ internal class TrieNode<E>(
...
+                            intersectionSizeRef += 1
+                            NOVALUE
+                        } else thisCell
+                    }
+                    // both are just E => compare them
+                    thisCell == otherNodeCell -> {
+                        intersectionSizeRef += 1
+                        NOVALUE
+                    }
+                    // both are just E, but different => return any one of them
```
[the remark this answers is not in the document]
belyaev-mikhail: Of course, this is a comment bug

belyaev-mikhail (commented):
> Why this test became useless?

It didn't, this is a merging issue, will fix.

core/commonMain/src/extensions.kt
```diff
@@ -291,6 +291,27 @@ operator fun <E> PersistentSet<E>.minus(elements: Array<out E>): PersistentSet<E
...
+ * or this instance if no modifications were made in the result of this operation.
+ */
+infix fun <E> PersistentSet<E>.intersect(elements: Iterable<E>): PersistentSet<E>
+        = if (elements is Collection) retainAll(elements) else mutate { it.retainAll(elements) }
+
+/**
+ * Returns all elements in this collection that are also
+ * contained in the specified [elements] collection.
+ *
+ * @return a new persistent collection with elements in this set that are also
```
qurbonzoda, line 309: return a new persistent ~collection~ **set**

core/commonMain/src/implementations/immutableSet/PersistentHashSet.kt
```diff
@@ -7,6 +7,7 @@ package kotlinx.collections.immutable.implementations.immutableSet
 
 import kotlinx.collections.immutable.PersistentSet
 import kotlinx.collections.immutable.internal.DeltaCounter
+import kotlinx.collections.immutable.internal.assert
```
qurbonzoda, line 10: Is it used in this file?

qurbonzoda (approved):
Great job!

qurbonzoda:
Please squash the commits

core/commonMain/src/implementations/immutableSet/PersistentHashSet.kt
belyaev-mikhail: Fixed

belyaev-mikhail:
Done @qurbonzoda

qurbonzoda:
🚀

Issue #92, opened 2020-10-30
Bug in PersistentList - list is broken after removeAll call


maio:
It only happens in certain situations. It can be reproduced using following test:

```
class PersistentListTest {
    @Test
    fun `persistent list fails`() {
        var xs = persistentListOf(
            *(1..1885).map { it }.toTypedArray()
        )

        xs = xs.removeAll(
            (1..1837).map { it }
        )

        println(xs.size)

        xs.forEach {
            print(it)
        }
    }
}
```

`print(it.i)` fails with exception 

```
java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class java.lang.Number
 ([Ljava.lang.Object; and java.lang.Number are in module java.base of loader 'bootstrap')
```

qurbonzoda:
Fixing commit is now in master and will part of the coming release.
Thank you for your bug report!

PR #94, opened 2020-11-19
Bug in PersistentList - list is broken after removeAll call #92

qurbonzoda:


ilya-g (approved):

core/commonMain/src/implementations/immutableList/PersistentVectorBuilder.kt
```diff
@@ -755,7 +755,7 @@ internal class PersistentVectorBuilder<E>(private var vector: PersistentList<E>,
         while (lastIndex shr rootShift == 0) {
             rootShift -= LOG_MAX_BUFFER_SIZE
             @Suppress("UNCHECKED_CAST")
-            newRoot = root[0] as Array<Any?>
+            newRoot = newRoot[0] as Array<Any?>
```
ilya-g, line 758: So usually it didn't happen because we only stepped once in this loop?

qurbonzoda: Yes, it appears none of the existing tests decrease root height

PR #102, opened 2021-03-23
Workaround specialized MutableEntrySet.contains implementation KT-41278

qurbonzoda:


core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentViews.kt
```diff
@@ -5,8 +5,18 @@
 
 package kotlinx.collections.immutable.implementations.immutableMap
 
+// intermediate abstract class to workaround KT-43321
+internal abstract class AbstractMapBuilderEntries<E : Map.Entry<K, V>, K, V> : AbstractMutableSet<E>() {
+    final override fun contains(element: E): Boolean {
+        // TODO: Eliminate this check after KT-30016 gets fixed.
+        if ((element as? Any?) !is Map.Entry<*, *>) return false
```
qurbonzoda, line 12: Interestingly js-ir doesn't prevent null element here. In case of null `element as Any?` leads to `ClassCastException`. Casting with `as?` fixes the exception.

TODO: Do some research to find out if this problem differs from KT-30016 and create a new issue if it does.

core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentViews.kt
```diff
@@ -21,16 +31,14 @@ internal class PersistentHashMapBuilderEntries<K, V>(private val builder: Persis
 
     override fun remove(element: MutableMap.MutableEntry<K, V>): Boolean {
```
qurbonzoda, line 32: TODO: Check if this function also prevents `Map.Entry`

qurbonzoda: Done: `BuilderEntries.remove` also needs the workaround to be able to remove `Map.Entry` elements

core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentViews.kt
ilya-g: Strange that we did without this check in stdlib. Do you have assertions in test that fail without it or fail with just `as Any?` cast?

qurbonzoda: This teamcity build failed with ClassCastException: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCollectionsImmutable_Build_All/3387162
The following commit added `as?` check and the tests passed: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCollectionsImmutable_Build_All/3387339

ilya-g: So, what test assertion causes such ClassCastException?

qurbonzoda: [This line](https://github.com/Kotlin/kotlinx.collections.immutable/blob/cd3c09d26a176ec2d98d5fd9476ed081ad06ca15/core/commonTest/src/contract/CollectionBehaviors.kt#L110) causes `contains` throw ClassCastException when `entries` is empty as the result of `clear` in [this line](https://github.com/Kotlin/kotlinx.collections.immutable/blob/cd3c09d26a176ec2d98d5fd9476ed081ad06ca15/core/commonTest/src/contract/map/ImmutableMapTest.kt#L246).
[This line](https://github.com/Kotlin/kotlinx.collections.immutable/blob/cd3c09d26a176ec2d98d5fd9476ed081ad06ca15/core/commonTest/src/contract/map/ImmutableMapTest.kt#L247) causes exception in `remove`.

ilya-g: > In case of null element as Any? leads to ClassCastException. 

Looks like an optimization based on the statically known non-null constraint of E.

ilya-g: I've checked what happens in stdlib, it's just that typecheck happens in  `containsEntry` implementation of each inheritor.

ilya-g (approved):

PR #104, opened 2021-03-31
Optimized equals() for all maps and sets

belyaev-mikhail:
Same as with other bulk operations, equals() can be streamlined to exploit HAMT/CHAMP internal structure for all set/map implementations.
Disclaimer: this optimization only works for two sets/maps of equal size using the same implementation and works best when equals will return false, which is already kind of a stretch, but we have cases for this exact setup.

Benchmarks show good results even for equal sets/maps:
https://docs.google.com/spreadsheets/d/1Y5N_-4S7yvEADMVtn82yP1bIP2wQ3mxXV3X1EL16Oes/edit?usp=sharing

Another questionable thing in this PR is support for comparing different implementations of maps (say, comparing an ordered map builder with an immutable hash map), which is mostly added because it's possible and not because it's realistic.

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
@@ -104,6 +104,12 @@ internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
         }
     }
 
+    /**
+     * We provide [equals], so as a matter of style, we should also provide [hashCode].
+     * However, the implementation from [AbstractMap] is enough.
+     */
+    override fun hashCode(): Int = super<AbstractMap>.hashCode()
+
     internal companion object {
         private val EMPTY = PersistentHashMap(TrieNode.EMPTY, 0)
         @Suppress("UNCHECKED_CAST")
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
@@ -8,6 +8,8 @@ package kotlinx.collections.immutable.implementations.immutableMap
 import kotlinx.collections.immutable.ImmutableCollection
 import kotlinx.collections.immutable.ImmutableSet
 import kotlinx.collections.immutable.PersistentMap
+import kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMap
+import kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilder
 import kotlinx.collections.immutable.mutate
 
 internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
@@ -75,6 +77,28 @@ internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
 
     override fun builder(): PersistentHashMapBuilder<K, V> {
         return PersistentHashMapBuilder(this)
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return when (other) {
+            is PersistentOrderedMap<*, *> -> {
+                node.equalsWith(other.hashMap.node) { a, b ->
+                    a == b.value
+                }
+            }
+            is PersistentOrderedMapBuilder<*, *> -> {
+                node.equalsWith(other.hashMapBuilder.node) { a, b ->
+                    a == b.value
+                }
+            }
+            is PersistentHashMap<*, *> -> {
+                node.equalsWith(other.node) { a, b -> a == b }
+            }
+            is PersistentHashMapBuilder<*, *> -> {
+                node.equalsWith(other.node) { a, b -> a == b }
+            }
+            else -> super.equals(other)
+        }
     }
 
     internal companion object {
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
@@ -80,6 +80,9 @@ internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
     }
 
     override fun equals(other: Any?): Boolean {
+        if (other === this) return true
+        if (other !is Map<*, *>) return false
+        if (size != other.size) return false
         return when (other) {
             is PersistentOrderedMap<*, *> -> {
                 node.equalsWith(other.hashMap.node) { a, b ->
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
@@ -130,7 +130,7 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
             }
 
             // should be super.equals(other), but https://youtrack.jetbrains.com/issue/KT-45673
-            else -> return other.entries.all { entry -> containsEntry(entry) }
+            else -> return other.all { entry -> containsEntry(entry) }
         }
 
     }
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
@@ -9,6 +9,7 @@ import kotlinx.collections.immutable.PersistentMap
 import kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMap
 import kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilder
 import kotlinx.collections.immutable.internal.DeltaCounter
+import kotlinx.collections.immutable.internal.MapImplementation
 import kotlinx.collections.immutable.internal.MutabilityOwnership
 
 internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap<K, V>) : PersistentMap.Builder<K, V>, AbstractMutableMap<K, V>() {
@@ -96,16 +97,6 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
         size = 0
     }
 
-    private fun <K1, V1> containsEntry(entry: Map.Entry<K1, V1>): Boolean {
-        entry is Map.Entry<*, *> || return false
-        val (k, v) = entry
-        val thisValue = get(k)
-        return when {
-            thisValue === null -> containsKey(k)
-            else -> thisValue == v
-        }
-    }
-
     override fun equals(other: Any?): Boolean {
         if (other === this) return true
         if (other !is Map<*, *>) return false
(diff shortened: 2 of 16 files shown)


qurbonzoda:
Can we add more `equals` tests, especially for maps/sets of different types?

qurbonzoda:
`hashCode` tests also needed.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -876,6 +876,36 @@ internal class TrieNode<K, V>(
...
+            if (buffer.size != that.buffer.size) return false
+            return (0 until buffer.size step ENTRY_SIZE).all { i ->
+                val thatKey = that.keyAtIndex(i)
+                val thatValue = that.valueAtKeyIndex(i)
+                (0 until buffer.size step ENTRY_SIZE).any { j ->
+                    val key = keyAtIndex(j)
+                    val value = valueAtKeyIndex(j)
+                    if (key == thatKey) {
+                        if (equalityComparator(value, thatValue)) true
+                        else return false
```
qurbonzoda, line 892:
```suggestion
                        return equalityComparator(value, thatValue)
```

belyaev-mikhail: This is not identical: `return false` here is a shortcut that returns from the whole function, while `true` is just the result of lambda passed to `any` =)

belyaev-mikhail:
> 
> 
> > Another questionable thing in this PR is support for comparing different implementations of maps (say, comparing an ordered map builder with an immutable hash map), which is mostly added because it's possible and not because it's realistic.
> 
> Here we can follow [stdlib definition of `equals` for maps](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-abstract-map/equals.html).

Well, we follow it by the spirit in all variants, but the question is whether it makes sense, for example, to optimize specifically for comparison between, say, ordered hashmap builder and unordered persistent hashmap. The chances this exact combination is ever needed by anyone are minimal, but it can be optimized in theory 'cos they follow the same tree structure.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
qurbonzoda: Oh, the suggestion is not an equivalent

belyaev-mikhail: Yep, i'd like to write it in a more obvious and elegant manner, but this code is already kinda' heavy on lambdas within lambdas and I just surrendered to this smelly variant.

qurbonzoda:
Seems like hashSet to hashSet and hashSet to orderedSet comparisions are not optimized. Maybe we should create an issue to track that, though hashSet and orderedSet are not based on the same tree structure.

belyaev-mikhail:
> Seems like hashSet to hashSet and hashSet to orderedSet comparisions are not optimized. Maybe we should create an issue to track that, though hashSet and orderedSet are not based on the same tree structure.

This specific case can be optimized as well in theory because the principal structure of the hash-based tree is still the same, but it needs a completely separate implementation specific for this particular case only, and a higher-order function like `equalsWith` would not help because of different underlying node classes. Not sure if it's worth it.

qurbonzoda:
> Well, we follow it by the spirit in all variants, but the question is whether it makes sense

Some combinations are really rare. We can easily omit the optimization for them if the implementation will be simplified cosiderably. If I got it right, the only simplification will be removing type checks in `when` block, which, I think, is not a big deal.

qurbonzoda:
hashSet to orderedSet comparison probably is not worth the effort for now, but we can record the issue as a known one.

It turns out hashSet to hashSet comparison is already optimized in `AbstractSet.equals`, which uses `containsAll`.

belyaev-mikhail:
> It turns out hashSet to hashSet comparison is already optimized in `AbstractSet.equals`, which uses `containsAll`

Yes, I had that in mind when optimizing `containsAll` which makes it the most important of the bulk operations in #90. Unfortunately, maps are not that lucky.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
qurbonzoda: You can cherry-pick this commit: https://github.com/Kotlin/kotlinx.collections.immutable/commit/72055f7d28b8fe5b915027ac946012d81f93a7e5

belyaev-mikhail: It seems you still removed the shortcut =)
```kotlin
                val keyIndex = collisionKeyIndex(thatKey)
                if (keyIndex != -1) {
                    val value = valueAtKeyIndex(keyIndex)
                    equalityComparator(value, thatValue)
                } else false
```
should still be
```kotlin
                val keyIndex = collisionKeyIndex(thatKey)
                if (keyIndex != -1) {
                    val value = valueAtKeyIndex(keyIndex)
                    equalityComparator(value, thatValue) || return false
                } else false
```

belyaev-mikhail:
Rebased onto master

core/commonMain/src/implementations/immutableMap/TrieNode.kt
qurbonzoda: Now that `any { }` function call is removed the `return false` is not a shortcut anymore.
Any `false` result will lead the enclosing `all { }` to return `false` immediately, the result of which is returned by `equalsWith`.

qurbonzoda:
Now that the `equals` function is overridden in those classes, IDEA inspections warn that `hashCode` is not overridden.
Maybe we can `@Suppress("EqualsOrHashCode")` the warning with comments? smth like "The `equals` function is overriden for optimization purposes and its behavior is not changed, so the inherided `hashCode` implementation is still valid."

core/commonTest/src/contract/map/ImmutableMapTest.kt
```diff
@@ -290,4 +291,53 @@ abstract class ImmutableMapTest {
...
+        assertFalse(base == persistentHashMapOf(*changed))
+        assertFalse(base == persistentMapOf(*changed))
+        assertFalse(base == persistentHashMapOf<ObjectWrapper<Int>, String>().builder().apply { putAll(changed) })
+        assertFalse(base == persistentMapOf<ObjectWrapper<Int>, String>().builder().apply { putAll(changed) })
+
+        val builder = immutableMapOf<ObjectWrapper<Int>, String>().builder().apply { putAll(data) }
+        assertTrue(builder == mapOf(*data))
+        assertTrue(builder == persistentHashMapOf(*data))
+        assertTrue(builder == persistentMapOf(*data))
+        assertTrue(builder == persistentHashMapOf<ObjectWrapper<Int>, String>().builder().apply { putAll(data) })
```
qurbonzoda, line 317: Minor: Because `ObjectWrapper<Int>` is used frequently in tests, we have declared a typealias for it - `IntWrapper`.

belyaev-mikhail:
> Maybe we can @Suppress("EqualsOrHashCode") the warning with comments? smth like "The equals function is overriden for optimization purposes and its behavior is not changed, so the inherided hashCode implementation is still valid."

I see two alternatives here: either suppress and write a comment, or just add a `hashCode` implementation that calls `super<AbstractMap>.hashCode` and add a comment right there.
As a personal preference I like the second approach better because in the first one the suppression and the comment will be at the top of the class, while in the second option it will be more local, but I'm not sure.

core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMap.kt
```diff
@@ -9,8 +9,9 @@ import kotlinx.collections.immutable.ImmutableCollection
 import kotlinx.collections.immutable.ImmutableSet
 import kotlinx.collections.immutable.PersistentMap
 import kotlinx.collections.immutable.implementations.immutableMap.PersistentHashMap
+import kotlinx.collections.immutable.implementations.immutableMap.PersistentHashMap.Companion.equals
```
qurbonzoda, line 12: Not sure why this is necessary

core/commonMain/src/implementations/immutableMap/PersistentHashMap.kt
```diff
@@ -104,6 +104,12 @@ internal class PersistentHashMap<K, V>(internal val node: TrieNode<K, V>,
         }
     }
 
+    /**
+     * We provide [equals], so as a matter of style, we should also provide [hashCode].
+     * However, the implementation from [AbstractMap] is enough.
+     */
+    override fun hashCode(): Int = super<AbstractMap>.hashCode()
```
qurbonzoda, line 111: Can we add a similar `hashCode` implementation to the affected builders as well?
Namely, `PersistentHashMapBuilder`, `PersistentOrderedMapBuilder`, `PersistentOrderedSetBuilder`.

core/commonMain/src/implementations/persistentOrderedSet/PersistentOrderedSet.kt
```diff
@@ -7,9 +7,8 @@ package kotlinx.collections.immutable.implementations.persistentOrderedSet
 
 import kotlinx.collections.immutable.PersistentSet
 import kotlinx.collections.immutable.implementations.immutableMap.PersistentHashMap
-import kotlinx.collections.immutable.implementations.persistentOrderedMap.LinkedValue
-import kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMap
-import kotlinx.collections.immutable.implementations.persistentOrderedMap.PersistentOrderedMapBuilder
+import kotlinx.collections.immutable.implementations.immutableMap.PersistentHashMap.Companion.equals
+import kotlinx.collections.immutable.implementations.immutableMap.PersistentHashMap.Companion.hashCode
```
qurbonzoda, line 11: Seems redundant

core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilder.kt
```diff
@@ -93,4 +95,43 @@ internal class PersistentHashMapBuilder<K, V>(private var map: PersistentHashMap
         node = TrieNode.EMPTY as TrieNode<K, V>
         size = 0
     }
+
+    private fun <K1, V1> containsEntry(entry: Map.Entry<K1, V1>): Boolean {
+        entry is Map.Entry<*, *> || return false
```
qurbonzoda, line 100: We handle the same issue [here](https://github.com/Kotlin/kotlinx.collections.immutable/blob/79cbf64d64b928868599e00f5629574be793e9dc/core/commonMain/src/implementations/immutableMap/PersistentHashMapContentViews.kt#L14) and in 3 more locations of the project. Maybe we can extract the handling to a function? Or at least have the implementation everywhere.

qurbonzoda: Hopefully, when the issue gets fixed or the legacy js backend is dropped, we can drop the workaround altogether.

qurbonzoda:
Other than the mentioned minor issues, the pull request LGTM.

core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMap.kt
belyaev-mikhail: This is some new IDEA quirk, I have no idea (pun not intended) why this is added or even why one would import a companion's method

qurbonzoda:
Merged manually.
Thank you!

PR #107, opened 2021-04-19
Add "JS IR supported" badge

SebastianAigner:
As part of our quest to increase awareness of the new Kotlin/JS IR compiler, we want to help users identify libraries which already provide artifacts for the JS/IR. This PR adds a badge to the README, which also links to the JS IR page in the documentation.


qurbonzoda:
Thank you!

PR #108, opened 2021-05-08
Fixed `toPersistentSet`, `toPersistentMap` to work with any impl.

comahe-de:
`toPersistentSet` and `toPersistentMap` now working in a similar way as `toPersistentList`

The contract in the KDoc was not fulfilled as it says for `toPersistentSet`:

> If the receiver is already a persistent set,
> returns it as is.
> If the receiver is a persistent set builder,
> calls `build` on it and returns the result.

But the previous implementation just checked for
`PersistentOrderedSet`, other implementations
of `PersistentSet` where not considered

Similar problem war fixed for `toPersistentMap`

### DIFF
--- a/core/commonMain/src/extensions.kt
+++ b/core/commonMain/src/extensions.kt
@@ -645,11 +645,14 @@ fun CharSequence.toImmutableSet(): PersistentSet<Char> = toPersistentSet()
  * If the receiver is already a persistent set, returns it as is.
  * If the receiver is a persistent set builder, calls `build` on it and returns the result.
  *
- * Elements of the returned set are iterated in the same order as in this collection.
+ * * If the receiver is not a persistent set builder,
+ * elements of the returned set are iterated in the same order as in this collection.
+ * Otherwise the order depends on the persistent set builder implementation.
+
  */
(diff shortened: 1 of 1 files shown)


qurbonzoda:
`toPersistentSet/Map` intensionally checks only for the ordered set. 
The idea here is to be consistent with [stdlib toMutableSet()](https://github.com/JetBrains/kotlin/blob/33b44585c43ee6746a33addfd14aa59c4bead6cb/libraries/stdlib/common/src/generated/_Collections.kt#L1692), which always returns an ordered set.

Can you please share your use case and why working with any implementation is a better behavior?

comahe-de:
@qurbonzoda I have some custom implementations of `PersistentList/-Set/-Map` that mainly delegate to the default `PersistentList/-Set/-Map` but add some additional properties. While a call to `toPersistentList()` on my `PersistentList` implementation  returns the same object, a call to `toPersistentSet()` on my `PersistentSet` implementation returns a new object, which was really unexpected as the doc says `If the receiver is already a persistent set, returns it as is.`

Regarding `toMutableSet()`: this method has a complete other behaviour: it always returns a new `Set` even if is already a `Set`. So the name is not well chosen.  `toNewMutableSet()` would be a better name. But this is a another topic.

Issue #109, opened 2021-05-09
PersistentHashMapBuilderTest.randomOperationTests fails on JVM


belyaev-mikhail:
Running full random tests on my machine fails with "The persistent map of 1 generation was modified. Expected: (...huge map here...) Actual: (...huge map here...)"

I believe this bug is there for some time (checked against my PRs, seems to fail on master back before #89).

belyaev-mikhail:
#110

PR #110, opened 2021-05-09
Fix for #109

belyaev-mikhail:
Seems like some logic here is broken: `mutableUpdateNodeAtIndex` expects `newNode` to be newly-created, but `mutableReplaceNode` does not check for it properly (`newNode` may be equal to `targetNode`, which may be any node of any builder previously created)


belyaev-mikhail:
Maybe the idea was to check `newNode.ownedBy` instead of `ownedBy`?
Anyway, removing the check fixes the test.

belyaev-mikhail:
I'd love to provide a regression test, but have no idea how to reproduce this outside the node implementation.
Even random test triggers this exclusively on JVM, because JVM implementation of these tests (and no other platform) actually calls builder's `remove(key, value)` method due to overloading witchcrafting, no other tests seem to call it under such stressful conditions.

PR #111, opened 2021-05-09
Fix for #109

belyaev-mikhail:
Github does not allow me to reopen #110, so a new PR

Comments from #110:

Seems like some logic here is broken: `mutableUpdateNodeAtIndex` expects `newNode` to be newly-created, but `mutableReplaceNode` does not check for it properly (`newNode` may be equal to `targetNode`, which may be any node of any builder previously created)

Maybe the idea was to check `newNode.ownedBy` instead of `ownedBy`?
Anyway, removing the check fixes the test.

I'd love to provide a regression test, but have no idea how to reproduce this outside the node implementation.
Even random test triggers this exclusively on JVM, because JVM implementation of these tests (and no other platform) actually calls builder's `remove(key, value)` method due to overloading witchcrafting, no other tests seem to call it under such stressful conditions.

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -784,10 +784,9 @@ internal class TrieNode<K, V>(
     private fun mutableReplaceNode(targetNode: TrieNode<K, V>, newNode: TrieNode<K, V>?, nodeIndex: Int, positionMask: Int, owner: MutabilityOwnership) = when {
         newNode == null ->
             mutableRemoveNodeAtIndex(nodeIndex, positionMask, owner)
-        ownedBy === owner || targetNode !== newNode ->
+        targetNode !== newNode ->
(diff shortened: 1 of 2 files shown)


qurbonzoda (commented):
You can observe the wrong behavior with the following test:
```
        val map0 = persistentHashMapOf<Int, Int>().put(0, 0).put(1, 1).put(32, 32)
        val map1 = map0.mutate { it.remove(0) }
        val map2 = map1.mutate {
            it.remove(1)
            it.remove(0)
        }

        assertTrue(map1.containsKey(32)) // fails
```
I am in favour of removing the `ownedBy === owner || ` check. It simplifies code a bit.
Let's add the regression test and we are done.

qurbonzoda:
An `assert(newNode.ownedBy === owner) // newNode is newly created by this builder` can also be added to check the assumption that if `newNode` is not the `targetNode`, than it must have been newly created.

belyaev-mikhail:
Rebased onto master + regression test (lousy naming) + assertion

qurbonzoda:
Thank you!

PR #116, opened 2021-07-15
PeristentHashMapBuilder.putAll() with another persistent hash map can produce incorrect results #114

qurbonzoda:
When putting all entries of a cell (1) into another cell (2), if (1) is an entry and (2) is a node, for optimization reasons the entry is put into the node. This leads to saving the old value of the entry if the node already contains the key.

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -460,38 +460,72 @@ internal class TrieNode<K, V>(
         }
     }
 
-    private fun mutablePutAllFromOtherNodeCell(other: TrieNode<K, V>,
-                                               positionMask: Int,
-                                               shift: Int,
-                                               intersectionCounter: DeltaCounter,
-                                               mutator: PersistentHashMapBuilder<K, V>): TrieNode<K, V> {
-        return when {
-            other.hasNodeAt(positionMask) -> {
-                mutablePutAll(
-                        other.nodeAtIndex(other.nodeIndex(positionMask)),
-                        shift + LOG_MAX_BRANCHING_FACTOR,
-                        intersectionCounter,
-                        mutator
-                )
-            }
-            other.hasEntryAt(positionMask) -> {
-                val keyIndex = other.entryKeyIndex(positionMask)
-                val key = other.keyAtIndex(keyIndex)
-                val value = other.valueAtKeyIndex(keyIndex)
-                val oldSize = mutator.size
-                val newNode = mutablePut(
-                        key.hashCode(),
-                        key,
-                        value,
-                        shift + LOG_MAX_BRANCHING_FACTOR,
-                        mutator
-                )
-                if (mutator.size == oldSize) {
-                    intersectionCounter.count++
+    /**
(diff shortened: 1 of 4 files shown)


core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -460,38 +460,72 @@ internal class TrieNode<K, V>(
...
+        }
+
+        otherNode.hasNodeAt(positionMask) -> {
+            val otherTargetNode = otherNode.nodeAtIndex(otherNode.nodeIndex(positionMask))
+            when {
+                this.hasEntryAt(positionMask) -> {
+                    // if otherTargetNode already has a value associated with the key, do not put this entry
+                    val keyIndex = this.entryKeyIndex(positionMask)
+                    val key = this.keyAtIndex(keyIndex)
+                    if (otherTargetNode.containsKey(key.hashCode(), key, shift + LOG_MAX_BRANCHING_FACTOR)) {
```
qurbonzoda, line 500: It is also possible to pass a flag, a la "do not update the associated value if the key already exists", to `mutablePut` to get rid of this `containsKey` check. However, the flag will leak to other functions from there and complicate the implementation. I think this is a fair price to pay for the sake of "simplicity". What are your thoughts about it?

qurbonzoda: Anyhow, I will run benchmarks to see how much it affects performance in certain scenarios.

chuckjaz:
FYI: I cherry-picked this into a CL and ran presubmit in the Compose tree and all tests pass. This doesn't appear to cause any semantic regressions but I run performance numbers in our CI so I have not checked performance.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
belyaev-mikhail: Summing up, looks good to me, and we have a clear improvement path in case it does introduce some significant overhead

belyaev-mikhail (approved):

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -586,7 +620,7 @@ internal class TrieNode<K, V>(
             else newNodeMap = newNodeMap or positionMask
             // we can use this later to skip calling equals() again
         }
-        assert(newNodeMap and newDataMap == 0)
+        check(newNodeMap and newDataMap == 0)
```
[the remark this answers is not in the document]
qurbonzoda: To be honest, when I started implementing these collections, I was a newcomer in Kotlin, so I didn't know there are `check` and `require` for assertion purposes ;)

core/commonMain/src/implementations/immutableMap/TrieNode.kt
qurbonzoda: The performance impact was within the error margin.

chuckjaz:
Thanks for the quick response to this!

FYI: A cherry-pick of this PR has landed in Compose main and b/193433239 is now marked as fixed.

PR #119, opened 2021-09-22
Update to Kotlin 1.5.31

### LINKED ISSUE #118: Support for M1 simulators
Since Kotlin 1.5.30 there are [M1 simulator targets available](https://kotlinlang.org/docs/whatsnew1530.html#apple-silicon-support). Would be great if you could add support for these as any project using this library cannot use the simulator targets until then. Hopefully it’s as simple as adding the new targets to build.gradle.kts.

ansman:
This also adds support for macosArm64.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,6 @@
 buildscript {
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.4.30")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.30")
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,6 @@
 buildscript {
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.30")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.31")
     }
 }
 
--- a/core/commonTest/src/stress/list/PersistentListBuilderTest.kt
+++ b/core/commonTest/src/stress/list/PersistentListBuilderTest.kt
@@ -622,7 +622,7 @@ class PersistentListBuilderTest : ExecutionTimeMeasuringTest() {
             vectorGen.add( builders.map { it.build() } )
             expected.add(lists)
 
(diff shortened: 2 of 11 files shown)


ansman:
I propose support for `iosSimulatorArm64`, `watchos` and `tvos` is also added.

build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 buildscript {
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.4.30")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.30")
```
JakeWharton, line 3: Might as well bump all the way to latest
```suggestion
        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.31")
```

ansman: I purposefully kept it low because Jetpack Compose only officially support 1.5.30

core/commonTest/src/stress/list/PersistentListBuilderTest.kt
```diff
@@ -622,7 +622,7 @@ class PersistentListBuilderTest : ExecutionTimeMeasuringTest() {
             vectorGen.add( builders.map { it.build() } )
             expected.add(lists)
 
-            val maxSize = builders.maxBy { it.size }?.size
+            val maxSize = builders.maxByOrNull { it.size }?.size
```
JakeWharton, line 625:
```suggestion
            val maxSize = builders.maxOf { it.size }
```
and others...

build.gradle.kts
ansman: But I guess bumping it here would only affect the library dependency which should be fine

mcpiroman: Jetpack Compose does now support 1.5.31

mcpiroman:
Second, preferably together with release :)

ZevEisenberg:
Bump - is this nearing release? This (three dependencies deep) is the last thing keeping me running Xcode in Rosetta mode on my M1 Mac.

qurbonzoda:
I have manually moved your changes to [another PR with Kotlin 1.6.0 and additional Native targets](https://github.com/Kotlin/kotlinx.collections.immutable/pull/121). The changes have already been released with the 0.3.5 version of the library.
Thank you for the effort!

PR #133, opened 2022-11-08
Optimize SmallPersistentVector.removeAll

mcpiroman:
Don't copy new buffer when removing single element.
Don't copy unecessary elements from old buffer to the new one.

I'm concerned that now the newBuffer can escape, so it may not be stack allocated.


mcpiroman:
Closed in favor of #134

PR #135, opened 2022-11-08
Bump kotlin, gradle, benchmarks

mcpiroman:


ilya-g:
Could you clarify why this change is needed?

mcpiroman:
I made it locally to streamline my development, but I don't think this is *needed*.

ilya-g:
Dependency update will happen in PR #148

PR #136, opened 2022-11-30
Don't allocate temporary buffer in SmallPersistentVector.removeAll

mcpiroman:
Actually we can avoid the buffer altogether and use 32-bit mask to track retained/removed elements.

Alternative to #134 

Again, I don't know which one is faster as I don't run the actual benchmarks

### DIFF
--- a/core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt
+++ b/core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt
@@ -49,30 +49,39 @@ internal class SmallPersistentVector<E>(private val buffer: Array<Any?>) : Immut
     }
 
     override fun removeAll(predicate: (E) -> Boolean): PersistentList<E> {
-        var newBuffer = buffer
-        var newSize = size
+        assert(MAX_BUFFER_SIZE <= Int.SIZE_BITS)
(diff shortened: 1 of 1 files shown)


core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt
```diff
@@ -49,30 +49,39 @@ internal class SmallPersistentVector<E>(private val buffer: Array<Any?>) : Immut
...
+
         return when (newSize) {
             size -> this
             0 -> EMPTY
-            else -> SmallPersistentVector(newBuffer.copyOfRange(0, newSize))
+            else -> {
+                // Align the mask to the left
+                retainedMask = retainedMask shl (Int.SIZE_BITS - size)
+
+                val newBuffer = bufferOfSize(newSize)
```
ilya-g, line 74: It could probably make sense to create the new buffer as `buffer.copyOf(newSize)` to avoid zeroing elements in the newly created buffer and then skip first retained elements there. But that's a guess as well, needs to be benchmarked.

qurbonzoda:
Moved review here: https://github.com/Kotlin/kotlinx.collections.immutable/pull/164

PR #143, opened 2023-03-21
Support Kotlin Community Build parameters

qurbonzoda:
The project takes the same set of Gradle parameters as the [kotlinx-benchmark](https://github.com/Kotlin/kotlinx-benchmark) project. Namely, `kotlin_version` specifies the Kotlin version to use, `kotlin_repo_url` specifies the repository to download artifacts from.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,14 +24,13 @@ infra {
 }
 
 allprojects {
-    logger.info("Using Kotlin $kotlin_version for project $this")
     repositories {
         addDevRepositoryIfEnabled(this, project)
         mavenCentral()
     }
 
-    // TODO: enable after https://youtrack.jetbrains.com/issue/KT-46257 gets fixed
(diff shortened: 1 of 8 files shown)


build.gradle.kts
```diff
@@ -21,7 +24,9 @@ infra {
 }
 
 allprojects {
+    logger.info("Using Kotlin $kotlin_version for project $this")
```
stasjas, line 27: please pass the compiler flag -version to log it

stasjas (approved):

build.gradle.kts
qurbonzoda: Done

stasjas (commented):
Let's postpone this changes as there is another solution (discussed in DM)

qurbonzoda:
The parameters support will be added here: https://github.com/Kotlin/kotlinx.collections.immutable/pull/194

PR #156, opened 2023-11-15
Wasm targets implementation

igoriakovlev:
Implement immutable collections for wasmJs and wasmWasi targets

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,8 @@
+import org.jetbrains.kotlin.gradle.dsl.KotlinJsCompile
+
 buildscript {
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.0")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.21")
     }
 }
 
@@ -26,5 +28,13 @@ allprojects {
 
     tasks.withType<org.jetbrains.kotlin.gradle.dsl.KotlinCompile<*>> {
         kotlinOptions.allWarningsAsErrors = true
+        kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
+        if (this is KotlinJsCompile) {
+            kotlinOptions.freeCompilerArgs += "-Xwasm-enable-array-range-checks"
+        }
     }
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask>().configureEach {
+    args.add("--ignore-engines")
 }
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -14,35 +14,34 @@ mavenPublicationsPom {
 }
 
 kotlin {
-    infra {
-        // According to https://kotlinlang.org/docs/native-target-support.html
+    applyDefaultHierarchyTemplate()
 
-        // Tier 1
-        target("linuxX64")
-        target("macosX64")
-        target("macosArm64")
-        target("iosSimulatorArm64")
-        target("iosX64")
+    // According to https://kotlinlang.org/docs/native-target-support.html
+    // Tier 1
+    linuxX64()
+    macosX64()
+    macosArm64()
+    iosSimulatorArm64()
+    iosX64()
 
-        // Tier 2
-        target("linuxArm64")
-        target("watchosSimulatorArm64")
-        target("watchosX64")
-        target("watchosArm32")
-        target("watchosArm64")
-        target("tvosSimulatorArm64")
-        target("tvosX64")
-        target("tvosArm64")
-        target("iosArm64")
+    // Tier 2
+    linuxArm64()
+    watchosSimulatorArm64()
+    watchosX64()
+    watchosArm32()
+    watchosArm64()
+    tvosSimulatorArm64()
+    tvosX64()
+    tvosArm64()
+    iosArm64()
 
-        // Tier 3
-        target("androidNativeArm32")
-        target("androidNativeArm64")
-        target("androidNativeX86")
-        target("androidNativeX64")
-        target("mingwX64")
-        target("watchosDeviceArm64")
-    }
+    // Tier 3
+    androidNativeArm32()
+    androidNativeArm64()
+    androidNativeX86()
+    androidNativeX64()
+    mingwX64()
+    watchosDeviceArm64()
 
     jvm {
         compilations.all {
@@ -65,6 +64,26 @@ kotlin {
                 sourceMap = true
                 moduleKind = "umd"
                 metaInfo = true
+            }
+        }
+    }
+
+    wasmJs {
+        nodejs {
+            testTask {
+                useMocha {
+                    timeout = "30000"
+                }
+            }
+        }
+    }
+
+    wasmWasi {
+        nodejs {
+            testTask {
+                useMocha {
+                    timeout = "30000"
+                }
             }
         }
     }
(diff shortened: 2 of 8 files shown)


ilya-g (changes requested):
Avoid duplicating sources for wasm target. Introduce a shared source set instead.

core/build.gradle.kts
```diff
@@ -14,34 +14,34 @@ mavenPublicationsPom {
 }
 
 kotlin {
-    infra {
-        // According to https://kotlinlang.org/docs/native-target-support.html
+    applyDefaultHierarchyTemplate {
```
igoriakovlev, line 17: Please pay attention to this update.

build.gradle.kts
```diff
@@ -26,5 +26,11 @@ allprojects {
 
     tasks.withType<org.jetbrains.kotlin.gradle.dsl.KotlinCompile<*>> {
         kotlinOptions.allWarningsAsErrors = true
+        kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
+        kotlinOptions.freeCompilerArgs += "-Xwasm-enable-array-range-checks"
```
ilya-g, line 30: Better to provide this option only to tasks that support it

core/build.gradle.kts
```diff
@@ -14,34 +14,34 @@ mavenPublicationsPom {
...
 
+        // According to https://kotlinlang.org/docs/native-target-support.html
         // Tier 1
-        target("linuxX64")
-        target("macosX64")
-        target("macosArm64")
-        target("iosSimulatorArm64")
-        target("iosX64")
+        this@kotlin.linuxX64()
```
ilya-g, line 21: Why explicit `this`?

core/build.gradle.kts
```diff
@@ -14,34 +14,34 @@ mavenPublicationsPom {
...
-        target("androidNativeX86")
-        target("androidNativeX64")
-        target("mingwX64")
-        target("watchosDeviceArm64")
+        this@kotlin.mingwX64()
+        this@kotlin.androidNativeArm32()
+        this@kotlin.androidNativeArm64()
+        this@kotlin.androidNativeX86()
+        this@kotlin.androidNativeX64()
+        this@kotlin.watchosDeviceArm64()
```
ilya-g, line 44: Can you preserve the order of targets?

core/build.gradle.kts
igoriakovlev: If no `this` fires `KotlinNativeTargetsWithHostTests can't be called in this context by implicit receiver. Use the explicit one if necessary`

build.gradle.kts
igoriakovlev: I put this argument for KotlinJsCompile task. There seems for now there is no a good way to put it into Wasm only compilation.

core/build.gradle.kts
ilya-g: Can you declare targets outside of `applyDefaultHierarchyTemplate` and call the latter without parameters, as it is suggested in docs? https://kotlinlang.org/docs/multiplatform-hierarchy.html#default-hierarchy-template

core/commonTest/src/contract/map/ImmutableMapTest.kt
```diff
@@ -283,13 +283,16 @@ abstract class ImmutableMapTest {
     }
 
     @Test fun noOperation() {
-        immutableMapOf<String, String>().toPersistentMap().testNoOperation({ clear() }, { clear() })
+        immutableMapOf<Any, String>().toPersistentMap().testNoOperation({ clear() }, { clear() })
```
ilya-g, line 286: Perhaps this change of the empty map types was not needed to fix the test

ilya-g:
Note that `:benchmarks:mingwX64BenchmarkGenerate` task fails when executing `assemble` task after updating Kotlin to 1.9.20. Fortunately, it isn't executed on TC in build and deploy configurations, so it doesn't block the build.

I presume this needs to be fixed in `kotlinx-benchmark` when updating it to Kotlin 1.9.20 
cc @qurbonzoda

core/commonTest/src/contract/map/ImmutableMapTest.kt
```diff
@@ -283,13 +283,16 @@ abstract class ImmutableMapTest {
...
-        val map = immutableMapOf("x" to 1, null to "x").toPersistentMap()
+        val instance = Any()
+        val instance2 = Any()
+
+        val map = immutableMapOf("x" to instance, null to "x").toPersistentMap()
         with(map) {
             testNoOperation({ remove("y") }, { remove("y") })
-            testNoOperation({ remove("x", 2) }, { remove("x", 2) })
-            testNoOperation({ put("x", 1) }, { put("x", 1) })     // does not hold
+            testNoOperation({ remove("x", instance2) }, { remove("x", instance2) })
```
qurbonzoda, line 294: Is it because Wasm does not have a pool of int objects?

core/commonTest/src/contract/map/ImmutableMapTest.kt
```diff
@@ -283,13 +283,16 @@ abstract class ImmutableMapTest {
...
+            testNoOperation({ put("x", instance) }, { put("x", instance) })     // does not hold
```
qurbonzoda, line 295: Does the test fail because Wasm does not have a pool of int objects?

core/commonTest/src/contract/map/ImmutableMapTest.kt
```diff
@@ -283,13 +283,16 @@ abstract class ImmutableMapTest {
...
             testNoOperation({ putAll(this) }, { putAll(this) })   // does not hold
```
qurbonzoda, line 296: Could you please also remove the `// does not hold` comments? The checks should pass in all platforms now.

core/commonTest/src/contract/map/ImmutableMapTest.kt
qurbonzoda, line 294: Weird: The `remove` provided for builder, is actually called on `map`.

qurbonzoda: Makes sense to eliminate the use of `with` scope function

qurbonzoda: I would suggest the following changes to the test: https://github.com/Kotlin/kotlinx.collections.immutable/commit/f3e3c61058be51937fc5e8c922ab93587607abb6
WDYT?

core/commonTest/src/contract/map/ImmutableMapTest.kt
igoriakovlev: Yes. Exactly.

core/build.gradle.kts
igoriakovlev: Fixed

core/build.gradle.kts
```diff
@@ -14,35 +14,33 @@ mavenPublicationsPom {
 }
 
 kotlin {
-    infra {
-        // According to https://kotlinlang.org/docs/native-target-support.html
```
ilya-g, line 18: Please preserve this comment

gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=org.jetbrains.kotlinx
 version=0.4
 versionSuffix=SNAPSHOT
-
```
ilya-g, line 4: Is this flag still needed?

gradle.properties
```diff
@@ -1,5 +1,5 @@
...
+kotlin.mpp.applyDefaultHierarchyTemplate=false
```
ilya-g, line 4: Is this flag still needed?

core/build.gradle.kts
igoriakovlev: Preserved

core/commonTest/src/contract/map/ImmutableMapTest.kt
igoriakovlev: done

core/commonTest/src/contract/map/ImmutableMapTest.kt
igoriakovlev: Yep, reverted

gradle.properties
```diff
@@ -1,5 +1,4 @@
 group=org.jetbrains.kotlinx
 version=0.4
 versionSuffix=SNAPSHOT
-
```
ilya-g, line 4: It may make sense to remove this change, so that file is untouched by this PR

ilya-g (approved):

gradle.properties
ilya-g: Note that it still has changes

ilya-g (approved):

igoriakovlev:
Thank you for the review.

PR #166, opened 2023-12-29
Add persistent Collection builder functions

### LINKED ISSUE #137: Add `buildPersistentList { }`, `buildPersistentMap { }`, `buildPersistentSet { }` to match the builders in kotlin stdlib
Add `buildPersistentList { }`, `buildPersistentMap { }`, `buildPersistentSet { }` to match the [collection builders in kotlin stdlib](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-list.html)

I would imagine it would look something like this: 
```kotlin
fun <E> buildPersistentList(block: (PersistentList.Builder<E>) -> Unit): PersistentList<E> = 
    persistentListOf<E>().builder().apply(block).build() 
```

Goooler:
Closes #137.


Goooler:
Depends on #165.

qurbonzoda (changes requested):
Documentation for the functions should be amended. Otherwise, LGTM.

core/commonMain/src/extensions.kt
```diff
@@ -766,3 +766,61 @@ fun <K, V> Map<K, V>.toPersistentHashMap(): PersistentMap<K, V>
         = this as? PersistentHashMap
         ?: (this as? PersistentHashMapBuilder<K, V>)?.build()
         ?: PersistentHashMap.emptyOf<K, V>().putAll(this)
+
+/**
+ * Builds a new [PersistentList] by populating a [PersistentList.Builder] using the given [builderAction]
+ * and returning an immutable list with the same elements.
+ *
+ * The list passed as a receiver to the [builderAction] is valid only inside that function.
+ * Using it outside the function produces an unspecified behavior.
```
qurbonzoda, line 775: Builders in kotlinx.collections.immutable are capable of constructing a new persistent collection multiple times. However, I'm struggling to identify a compelling use case for retaining a reference to the builder and utilizing it outside the builderAction. If we constraint the builder to be valid only within the function, could using a `MutableList` as the receiver be a more performance-efficient choice?

qurbonzoda: cc @ilya-g, WDYT?

core/commonMain/src/extensions.kt
```diff
@@ -766,3 +770,76 @@ public fun <K, V> Map<K, V>.toPersistentHashMap(): PersistentMap<K, V>
...
+ */
+@OptIn(ExperimentalTypeInference::class, ExperimentalContracts::class)
+public inline fun <T> buildPersistentList(@BuilderInference builderAction: PersistentList.Builder<T>.() -> Unit): PersistentList<T> {
+        contract { callsInPlace(builderAction, InvocationKind.EXACTLY_ONCE) }
+        return persistentListOf<T>().builder().apply(builderAction).build()
+}
```
Goooler, line 785: 1. Annotate `@BuilderInference` for better generic type support.
2. Add `InvocationKind.EXACTLY_ONCE` contract to indicate  the function parameter will be invoked exactly one time.

Goooler, line 785: 1. Annotate `@BuilderInference` for better generic type support.
2. Add `InvocationKind.EXACTLY_ONCE` contract to indicate  the function parameter will be invoked exactly one time.

kyay10: AFAIK `@BuilderInference` is a no-op, and has been for a while now. My understanding is that builder inference is applied by default everywhere

core/commonMain/src/extensions.kt
```diff
@@ -766,3 +770,76 @@ public fun <K, V> Map<K, V>.toPersistentHashMap(): PersistentMap<K, V>
...
+ * Builds a new [PersistentList] by populating a [PersistentList.Builder] using the given [builderAction]
+ * and returning an immutable list with the same elements.
+ *
+ * The list passed as a receiver to the [builderAction] is valid only inside that function.
+ * Using it outside the function produces an unspecified behavior.
+ */
+@OptIn(ExperimentalTypeInference::class, ExperimentalContracts::class)
+public inline fun <T> buildPersistentList(@BuilderInference builderAction: PersistentList.Builder<T>.() -> Unit): PersistentList<T> {
+        contract { callsInPlace(builderAction, InvocationKind.EXACTLY_ONCE) }
+        return persistentListOf<T>().builder().apply(builderAction).build()
```
kyay10, line 784: Could use `mutate` instead. Same for the other builder functions

Goooler: Good catch! Replaced.

core/commonMain/src/extensions.kt
kyay10: It feels like if that were the case, `mutate` would already be optimizing this in the same way, no?

kyay10:
AFAIK, [builder inference](https://kotlinlang.org/docs/using-builders-with-builder-inference.html) no longer requires the `@BuilderInference` annotation. There's no harm in including it, but it's just a bit of code clutter.

core/commonMain/src/extensions.kt
Goooler: Removed now.

kyay10: The opt-in for `ExperimentalTypeInference::class` is unnecessary now

Goooler:
https://github.com/Kotlin/kotlinx.collections.immutable/blob/bd5ac10f1e3b72c8bd6335005bf27d751e0d8b52/core/commonMain/src/extensions.kt#L41

`builder` and `apply` are applied for `mutate`, and I see the contract is already on the `apply`:

```kt
@kotlin.internal.InlineOnly
public inline fun <T> T.apply(block: T.() -> Unit): T {
    contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }
    block()
    return this
}
```

core/commonMain/src/extensions.kt
Goooler: Removed.

Goooler:
CC @DmitryNekrasov

PR #169, opened 2024-01-05
Introduce binary compatibility validator

### LINKED ISSUE #168: Introduce Binary compatibility validator to track API changes
https://github.com/Kotlin/binary-compatibility-validator

Goooler:
Closes #168.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,6 +8,7 @@ buildscript {
 
 plugins {
     id("kotlinx.team.infra") version "0.4.0-dev-80"
+    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.13.2"
 }
 
 infra {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -21,6 +22,16 @@ infra {
     }
 }
 
+apiValidation {
+    ignoredPackages += listOf(
+        "kotlinx.collections.immutable.internal"
```
qurbonzoda, line 27: The package doesn't seem to have any public declaration currently. But I think it is important to catch unintended public declarations in this package in the future.

qurbonzoda (changes requested):
I would suggest validating public declarations in the `kotlinx.collections.immutable.internal` package as well. Otherwise, LGTM

build.gradle.kts
Goooler: Done.

qurbonzoda:
Thanks for the contribution!

PR #174, opened 2024-02-28
Add Kotlin K2 migrations

tbogdanova:
Building the project with the K2 version of the compiler revealed several migrations. 
It is necessary to move these changes to `master` branch to simplify the support of `kotlin-community/dev` branch. As well as simplifying the future migration of the project to a new version of the compiler.

### DIFF
--- a/core/commonTest/src/contract/list/ImmutableListTest.kt
+++ b/core/commonTest/src/contract/list/ImmutableListTest.kt
@@ -25,7 +25,7 @@ class ImmutableListTest {
 
         assertFailsWith<NoSuchElementException> { empty1.iterator().next() }
 
-        compareLists(emptyList(), empty1)
+        compareLists(emptyList<Int>(), empty1)
 
     }
 
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -121,8 +121,10 @@ kotlin {
         }
 
         val wasmMain by creating {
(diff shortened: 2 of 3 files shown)


core/commonTest/src/contract/list/ImmutableListTest.kt
```diff
@@ -25,7 +25,7 @@ class ImmutableListTest {
 
         assertFailsWith<NoSuchElementException> { empty1.iterator().next() }
 
-        compareLists(emptyList(), empty1)
+        compareLists(emptyList<Int>(), empty1)
```
ilya-g, line 28: AFAIK, this won't be a problem in K2 soon

tbogdanova: Warning will remain. See [KT-66196](https://youtrack.jetbrains.com/issue/KT-66196).

tbogdanova: But let's decide if we want to take this into the main branch.

qurbonzoda: Given that the change is needed for K2, I don't have objections to merging it into `master`. This also aligns with the goal of keeping the `kotlin-community/dev` branch close to `master`.

qurbonzoda (approved):

PR #175, opened 2024-03-01
feat: add jpms support with `module-info.java`

### LINKED ISSUE #157: Provide module-info descriptor for java 9 and later
It would be great if this library was provided as a multi-release jar so that it could be part of the Java module system.

sgammon:
## Summary

This PR adds a `module-info.java` definition, bringing `kotlinx-collections-immutable` to full JPMS support; downstream users can use the new module with the following directive:

**`module-info.java`**
```java
module your.module.here {
  requires kotlinx.collections.immutable;
}
```

Fixes and closes #157 

### Other changes

In general, the build was cleaned up, and Dependency Locking and Dependency Verification are now turned on, as is the case for other Kotlin projects.

### Trying it out

This is still a draft because it is being tested downstream. If you would like to test this module, you can do so with:

**`build.gradle.kts`**
```kotlin
dependencies {
  implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.3.8-rc-1")
}
```

**`settings.gradle.kts`**
```kotlin
dependencyResolutionManagement {
  repositories {
    mavenCentral()

    // 👇  this repo
    maven {
      name = "elide-snapshots"
      url = uri("https://elide-snapshots.storage-download.googleapis.com/repository/v3/")
    }
  }
}
```

## Changelog

- feat: add `module-info.java`
- feat: add `Java9Modularity` used by other kotlinx projects
- feat: build `core` project as multi-release jar
- chore: upgrade gradle → `8.6` (to run under jvm21)
- chore: upgrade kotlin → `1.9.22`
- chore: general build cleanup (use kotlin for most build scripts)
- chore: add gradle lockfiles and locking config
- chore: add gradle dependency verification config


benchmarks/runner/build.gradle
```diff
@@ -3,8 +3,8 @@ plugins {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib"
-    implementation 'org.openjdk.jmh:jmh-core:1.21'
+    implementation libs.kotlin.stdlib
+    implementation libs.jmh.core
```
sgammon, line 7: General build cleanup: moved dependencies to a version catalog.

build.gradle.kts
```diff
@@ -45,4 +45,8 @@ allprojects {
 
 tasks.withType<org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask>().configureEach {
     args.add("--ignore-engines")
-}
\ No newline at end of file
+}
+
+dependencyLocking {
+    lockAllConfigurations()
+}
```
sgammon, line 52: Dependency locking and verification were activated, but kept in lenient mode for now, to account for native dependency shift in CI.

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
+import org.jetbrains.kotlin.tooling.core.*
+import java.io.*
+import kotlin.reflect.*
+import kotlin.reflect.full.*
+
+object Java9Modularity {
+    @JvmStatic
+    @JvmOverloads
+    fun Project.configureJava9ModuleInfo(multiRelease: Boolean = true) {
```
sgammon, line 33: `Java9Modularity` added from [here](https://github.com/Kotlin/kotlinx.serialization/blob/master/buildSrc/src/main/kotlin/Java9Modularity.kt)

core/build.gradle.kts
```diff
@@ -6,7 +10,7 @@ plugins {
 }
 
 base {
-    archivesBaseName = "kotlinx-collections-immutable" // doesn't work
+    archivesName = "kotlinx-collections-immutable"
```
sgammon, line 13: Fixes deprecation

core/build.gradle.kts
```diff
@@ -1,3 +1,7 @@
+@file:OptIn(org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl::class)
```
sgammon, line 1: Fixes warnings

core/jvmMainModule/src/module-info.java
```diff
@@ -0,0 +1,4 @@
+module kotlinx.collections.immutable {
+    requires java.base;
+    requires transitive kotlin.stdlib;
+}
```
sgammon, line 4: New `module-info.java` descriptor. I thought about adding it as `kotlinx.collections.immutable.jvm`, which is the suggested module name in IntelliJ; but I don't think users are using this value yet, because I can't locate it anywhere as an `Automatic-Module-Name`.

Thus, the format here matches other modules, like:
- `kotlinx.coroutines.[core|guava|...]`
- `kotlinx.serialization.[core|json|...]`

gradle/libs.versions.toml
```diff
@@ -0,0 +1,20 @@
+[versions]
+kotlin-sdk = "1.9.22"
+kotlinx-benchmark = "0.4.10"
+kotlinx-validator = "0.13.2"
+guava-testlib = "18.0"
+jmh = "1.21"
```
sgammon, line 6: New version catalog; other than updating `kotlin-sdk` to `1.9.22`, no version changes were applied.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-bin.zip
```
sgammon, line 3: Gradle upgrade → latest stable so that this project can run under JVM 21.

settings.gradle.kts
```diff
@@ -0,0 +1,14 @@
+pluginManagement {
+    repositories {
```
sgammon, line 2: Rewrote this file as Kotlin because it helps a bit with build speed, since Groovy build scripts can't be build cached until Gradle 8.7, which is at RC2 (not yet released).

sgammon:
cc / @qurbonzoda @daniel-jasinski @Sporking @qwwdfsad

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+import java.io.FileInputStream
+
+plugins {
+    `kotlin-dsl`
+}
+
+repositories {
+    mavenCentral()
+    mavenLocal()
+    if (project.hasProperty("kotlin_repo_url")) {
```
daniel-jasinski, line 16: Use the nullable `findProperty` method to check and retrieve property. The `properties` property performs a copy of the underlying map.

daniel-jasinski:
Good PR, but there is a lot of housekeeping and the original intention gets lost in all the minutiae.
I would suggest splitting this PR, at least into the **chore** and **feature** parts.

ilya-g (changes requested):
While this project contains only one module where module-info should be added, it would be better to avoid introducing an included build with common logic like `buildSrc` - it complicates build and increases configuration time.

It's enough to add a compilation task directly in the `core` subproject, like it is done in [kotlinx.datetime](https://github.com/Kotlin/kotlinx-datetime/blob/v0.5.0/core/build.gradle.kts#L258)

We can refactor it later to shared build logic if we introduce another published module.

sgammon:
@daniel-jasinski Thank you for that review

@ilya-g Okay, no problem. I will apply those changes.

buildSrc/build.gradle.kts
sgammon: Good catch, although this block wasn't supposed to make it in at all; I'll drop

DmitryNekrasov:
Superseded by #268, which adds the minimal module-info.java agreed in #157. Closing. Thanks for starting this off.

sgammon:
thanks for picking this up @DmitryNekrasov 👍

PR #176, opened 2024-03-12
Avoid creating new PersistentList instance when adding empty collection

popematt:
### Problem

I encountered unexpected growth in memory usage while working on an library that uses `PersistentList`. I eventually realized that it is because I had many distinct instances of an empty `PersistentList`.

I traced down the problem, starting with `toPersistentList()`, which I was using to convert all inputs before creating assembling it into my data model. The `toPersistentList()` function takes the empty persistent list and adds all elements of `this` to the empty list. When you add elements to a `SmallPersistentVector`, it checks whether the combined size is less than or equal to `MAX_BUFFER_SIZE` and if it is, it unconditionally creates a new `SmallPersistentVector` instance.

https://github.com/Kotlin/kotlinx.collections.immutable/blob/7fb0d74ea87d1f52e8d30d39d4df066a57f51e50/core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt#L38-L49

This means that no-ops on small lists (such as adding two empty lists) always results in a new instance being created.

I verified that the problem only exists for `SmallPersistentVector`. `PersistentVector` uses `PersistentVectorBuilder`, which already has checks for adding an empty collection.

### What I am changing

I have added a check to `fun addAll(elements: Collection<E>): PersistentList<E>` and `fun addAll(index: Int, c: Collection<E>): PersistentList<E>` to return early if the collection of elements to be added is empty.

I have updated the tests to include missing no-op cases related to adding and removing empty collections in both `ImmutableListTest` and `ImmutableSetTest`, and I have added tests for `toPersistentMap()`, `toPersistentSet()`, and `toPersistentList()` to ensure that when the receiver is an empty map/collection, they always return the same instance of persistent map/set/list.

### DIFF
--- a/core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt
+++ b/core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt
@@ -36,6 +36,7 @@ internal class SmallPersistentVector<E>(private val buffer: Array<Any?>) : Immut
     }
 
     override fun addAll(elements: Collection<E>): PersistentList<E> {
+        if (elements.isEmpty()) return this
         if (size + elements.size <= MAX_BUFFER_SIZE) {
             val newBuffer = buffer.copyOf(size + elements.size)
             // TODO: investigate performance of elements.toArray + copyInto
@@ -80,6 +81,7 @@ internal class SmallPersistentVector<E>(private val buffer: Array<Any?>) : Immut
 
     override fun addAll(index: Int, c: Collection<E>): PersistentList<E> {
         checkPositionIndex(index, size)
+        if (c.isEmpty()) return this
         if (size + c.size <= MAX_BUFFER_SIZE) {
             val newBuffer = bufferOfSize(size + c.size)
             buffer.copyInto(newBuffer, endIndex = index)
--- a/core/commonTest/src/contract/list/ImmutableListTest.kt
+++ b/core/commonTest/src/contract/list/ImmutableListTest.kt
@@ -68,6 +68,13 @@ class ImmutableListTest {
         compareLists(list, immList)
     }
 
+    @Test fun emptyListToPersistentList() {
+        val empty = emptyList<Int>()
+        val emptyPersistent = empty.toPersistentList()
+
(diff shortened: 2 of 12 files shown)


core/commonTest/src/contract/list/ImmutableListTest.kt
```diff
@@ -68,6 +68,13 @@ class ImmutableListTest {
         compareLists(list, immList)
     }
 
+    @Test fun emptyListToPersistentList() {
+        val empty = emptyList<Int>()
+        val emptyPersistent = empty.toPersistentList()
+
+        assertTrue(emptyPersistent === empty.toPersistentList())
```
JakeWharton, line 75: There is `assertSame` for this, which should give a _slightly_ better error message should it ever fail.

popematt: Yeah, my preference is to use `assertSame()` but I noticed that this particular file was using `===` instead of `assertSame()`, and I try not to introduce new things for small changes like this. I'm happy to change it, though.

JakeWharton: Makes sense. It probably pre-dates instance equality being added to kotlin.test! No need to change, was just making sure it was known.

JakeWharton (commented):
I'm not a maintainer, so best I can do is 👍

core/commonMain/src/implementations/immutableList/SmallPersistentVector.kt
```diff
@@ -36,6 +36,7 @@ internal class SmallPersistentVector<E>(private val buffer: Array<Any?>) : Immut
     }
 
     override fun addAll(elements: Collection<E>): PersistentList<E> {
+        if (elements.isEmpty()) return this
```
qurbonzoda, line 39: What do you think about adding the same check to other persistent collections? 
Even though they eventually return the same instance, they create a throwaway builder in the case of an empty argument.

qurbonzoda (changes requested):

qurbonzoda:
Thank you very much for the contribution!

PR #177, opened 2024-04-03
Add missing toPersistentMap, rename persistentXxxOf to emptyPersistentXxx

Laxystem:
I've encountered some annyoances when using KotlinX Coroutines, and so I've decided to move my own scattered-around solutions into the library itself.
* The functions to create empty collections/maps are named inconsistently with the stdlib; `persistentXxxOf` instead of `emptyXxx`.
* There are no functions to create persistent and immutable maps from sequences, arrays, and iterables of pairs. The only options are to call `list.toMap().toImmutableMap()` or `persistentMapOf<Xxx>().mutate { addAll(list) } `.
* Replacing `immutableXxxOf` with `persistentXxxOf` results in bugged code, as the type parameter is ignored.
* The library didn't build at all on my Linux laptop. I have added some gradle properties to fix that.

Though I do have some questions:
* Should error-deprecated `emptyImmutableXxx()` no-op functions be created, similarly to `Flow.forEach { }`?
* Should the `.toImmutableMap()` functions be removed or error-deprecated? As far as I understand, `Map.toImmutableMap` is kept as it returns the receiver if it is an instance of `ImmutableMap`; But the new functions simply call their persistent versions.
* As a side-note, should `kotlin-js-store/yarn.lock` be added to the `.gitignore`, or should it be committed?

Sorry for the big PR; Lots of the changes I did were in the same file and some even in the same lines, so it made more sense to submit them together than separately.


Laxystem:
@qurbonzoda are you intending on merging this as a part of the beta? locally, all tests ran; the changes to seemingly all files is just replacing usage of `persistentXxxOf` with `emptyXxx`; and the previous functions still exist, but are deprecated; in short, this scary PR really isn't that scary.

qurbonzoda:
> The functions to create empty collections/maps are named inconsistently with the stdlib; persistentXxxOf instead of emptyXxx.

The standard library has both `collectionOf()` and `emptyCollection()` variants. For example, [listOf()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/list-of.html) and [emptyList()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/empty-list.html). The same of `Set` and `Map`.
Hence, I suggest adding the `emptyPersistentCollection()` functions in addition to the existing `persistentCollectionOf()`, not instead of them.

qurbonzoda:
> should kotlin-js-store/yarn.lock be added to the .gitignore, or should it be committed?

Please feel free to commit it in a separate PR.

qurbonzoda:
> Replacing immutableXxxOf with persistentXxxOf results in bugged code, as the type parameter is ignored.

I can't reproduce this issue. Could you please clarify what IDE and its version you are using?

qurbonzoda:
> The library didn't build at all on my Linux laptop. I have added some gradle properties to fix that.

Could you please add them in a separate commit with description of the fix?

qurbonzoda:
> Should the .toImmutableMap() functions be removed or error-deprecated? As far as I understand, Map.toImmutableMap is kept as it returns the receiver if it is an instance of ImmutableMap; But the new functions simply call their persistent versions.

Calling the persistent version is an implementation detail. We could return a different collection that implements `ImmutableMap` but not `PersistentMap`, which might have a better performance.

qurbonzoda:
> Should error-deprecated emptyImmutableXxx() no-op functions be created, similarly to Flow.forEach { }?

Could you please clarify why this might be useful?

core/commonMain/src/extensions.kt
```diff
@@ -766,3 +806,107 @@ public fun <K, V> Map<K, V>.toPersistentHashMap(): PersistentMap<K, V>
...
         ?: PersistentHashMap.emptyOf<K, V>().putAll(this)
+
+
+/**
+ * Returns a persistent map of all elements in this [Sequence] of [Pair]s,
+ * using [Pair.first] as [keys][ImmutableMap.keys] and [Pair.second] as [values][ImmutableMap.values].
+ *
+ * @since 0.4
+ */
+public fun <T, R> Sequence<Pair<T, R>>.toPersistentMap(): PersistentMap<T, R> =
```
qurbonzoda, line 817: Could you please add tests for the new functions?

Laxystem: Not really; Idk how the current functions are tested, so I can't know how you want the tests to be done. I could not find these tests myself.

qurbonzoda: `Map` tests are mostly in this file: https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/core/commonTest/src/contract/map/ImmutableMapTest.kt

You can add tests for new functions either there or in a new file within the directory.

qurbonzoda:
>> Could you please add them in a separate commit with description of the fix?

> Well no, I've already committed

You can always use Git commands, such as reset and interactive rebase, to reorganize your commits. Please note that for the PR to be merged, each separate change should be in a separate commit. For example, there should be one commit for introducing the `emptyPersistentCollection()` functions and another for introducing the `toPersistentMap/HashMap()` functions.

If you find reorganizing commits difficult, please let me know. I can do it before merging.

benchmarks/commonMain/src/benchmarks/immutableList/AddAll.kt
```diff
@@ -35,7 +36,7 @@ open class AddAll {
      */
     @Benchmark
     fun addAllLast(): ImmutableList<String> {
-        return persistentListOf<String>().addAll(listToAdd)
+        return emptyPersistentList<String>().addAll(listToAdd)
```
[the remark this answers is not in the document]
Laxystem: I could; Isn't it clearer this way, however?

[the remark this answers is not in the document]
qurbonzoda: I believe it's a matter of preference, and not worth modifying a lot of files.

Laxystem:
@qurbonzoda could you please implement all the changes you've detailed? I'm sorry, but I can't find time to this myself—I'd love to, sorry.

PR #178, opened 2024-04-08
Upgraded Gradle to 8.7

severn-everett:
Added checksum configuration in Gradle Wrapper properties; Converted settings.gradle to use Kotlin DSL


core/build.gradle.kts
```diff
@@ -1,12 +1,13 @@
...
+import org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl
 
 plugins {
     id("kotlin-multiplatform")
     `maven-publish`
 }
 
 base {
-    archivesBaseName = "kotlinx-collections-immutable" // doesn't work
+    archivesName.set("kotlinx-collections-immutable") // doesn't work
```
Goooler, line 10:
```suggestion
    archivesBaseName = "kotlinx-collections-immutable" // doesn't work
```

This assignment could be used on newer Gradle.

https://blog.gradle.org/simpler-kotlin-dsl-property-assignment

PR #180, opened 2024-04-25
Enable KLib ABI validation

Goooler:
https://github.com/Kotlin/binary-compatibility-validator?tab=readme-ov-file#experimental-klib-abi-validation-support

Follow up #169.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
```
Goooler, line 3: Newer BCV requires newer Gradle.

fzhinkin: That's a bug in BCV. 😞 

https://github.com/Kotlin/binary-compatibility-validator/pull/204 should fix it.

Goooler: Good to know! But anyway, it's worth to bump Gradle, the current using one is so old.

Issue #181, opened 2024-05-14
Map, Filter, etc. operations on an ImmutableList should return another ImmutableList


nick2525:
I am writing to share suggestions regarding the use of `ImmutableList` within the Kotlin ecosystem, particularly in the context of Android development using Jetpack Compose. As a Kotlin developer, I have greatly appreciated the language's pragmatic approach and its powerful features that generally promote a smooth development experience. However, I've encountered a specific issue with `ImmutableList` that I believe warrants attention to better align with Kotlin's philosophy and the needs of modern Android development.

While working with Jetpack Compose, the immutability of data structures is paramount to optimize UI updates and performance. Kotlin's `ImmutableList` initially seemed like a fitting choice to guarantee data integrity and promote efficient recompositions. However, a significant drawback arises from the fact that transformations on `ImmutableList`, such as using the `map` function, return a standard mutable `List` (backed by `ArrayList` under the hood). This behavior not only contradicts the expectations for an immutable collection but also forces developers to frequently convert results back to `ImmutableList` using `toImmutableList()`. This pattern results in excessive object creation, which is counterproductive and inefficient, particularly in a performance-sensitive environment like mobile development.

Given Kotlin's commitment to being a pragmatic language, the current handling of `ImmutableList` feels somewhat misaligned with this vision. The necessity to revert to mutable types and then convert them back to immutable ones is reminiscent of older Java practices, which Kotlin has otherwise superbly improved upon.

To address this, I propose the introduction of a more integrated approach for immutable collections in Kotlin, particularly for transformations like `map`, `filter`, etc. Ideally, these operations on an `ImmutableList` should return another `ImmutableList` directly, without intermediate mutable stages. This enhancement would not only reduce overhead from unnecessary object creation but also streamline development by adhering more closely to functional programming principles, where immutability is a standard expectation. Also, I don't want to create ArrayList. I want there to be complete immutability in the application.

Implementing this change could significantly improve Kotlin’s utility and efficiency in Android development, particularly when using Jetpack Compose.

Also, the implementation of Map, Filter, operations on an ImmutableList can be made as a separate library, and I really want there to be no ArrayList under the hood

also reference: https://youtrack.jetbrains.com/issue/KT-68071#focus=Comments-27-9803975.0-0

qurbonzoda:
Hi!
Thanks for the feedback!
Indeed this is an annoying issue and we already have a couple of related issues.
[...]

Issue #10, opened 2017-03-16
Importing operators


cbruegg:
From the readme:
> Note: you need to import these operators from kotlinx.collections.immutable package in order for them to take the precedence over the ones from the standard library.

That makes it very easy to accidentally use the standard operators, which obviously has bad performance implications. Perhaps the operator functions should be instance methods so that the compiler prefers them over the standard `plus`/`minus` extension functions?

ilya-g:
Yes, that is a valid concern. 

Unfortunately making them member would clutter the interfaces.

I have another option in mind: if the standard operators know something about immutable collections, they can be specialized in runtime to utilize immutability of those collections.

ilya-g:
We have discussed the issue. 

There were three options how to resolve it:

1. Name all operations differently
    - good names are not cheap
    - operators can't be renamed
    - it introduces even more confusion

2. Make extensions on immutable collections more preferable than the same-named extensions on read-only collections, even when they're not imported.

    2.1. Light way: make completion prefer package with those extensions and insert required import automatically.
    - removing import changes the behavior and it can be left unnoticed, copy-pasting code block without imports changes the behavior.
 
    2.2. Hard way: allow to distinguish these extensions, so they are follow the type as members and do not require explicit import. 
    - may require a lot of design work in compiler

3. Make immutable collection interfaces not extend read-only collection interfaces at all, including `Iterable<T>`. 
    - This way no existing `Iterable/List/Set/Map` extensions  would be availalbe on immutable collections. 
    - In order to pass such collections where a read-only collection is requires, an explicit conversion, i.e. `asList()` or `asMap()`, should be invoked.
    - An immutable collection implementation may actually implement corresponding read-only collection, so that conversion will be a no-op.

Out of all options the third looks the most realistic, though on a first sight it might look awkward that an immutable collection do not extend read-only one.

What do you guys think about it?

cbruegg:
Option 3 sounds fair to me, especially since 1 and 2 really don't seem like feasible solutions. 
Would all of the current extensions to `Iterable/List/Set/Map` be made available separately and optimized for the corresponding immutable types? Most of my method calls where collections are passed are calls to those, so if they existed, there wouldn't be too many places where I'd need to add `asList()` in my application code.

Sporking:
[...]

PR #183, opened 2024-05-15
Add orEmpty extensions for immutable collections

### LINKED ISSUE #182: No orEmpty function for ImmutableList?
Can we add an extension function similar to `List` for `ImmutableList`?

```
@kotlin.internal.InlineOnly
public inline fun <T> ImmutableList<T>?.orEmpty(): List<T> = this ?: persistentListOf()
```

Goooler:
Closes #182.


qurbonzoda (approved):

st-hocnguyen (commented):
What about `PersistentList.orEmpty: PersistentList`, `PersistentMap.orEmpty: PersistentMap` and `PersistentSet.orEmpty: PersistentSet` overloads 🙏 ?

Goooler:
`Persistent*` extends `Immutable*`, these extensions work for them both.

st-hocnguyen:
> `Persistent*` extends `Immutable*`, these extensions work for both of them.

But when we need a value of `Persistent*`, we must do `orEmpty().toPersistent*()`. I think adding above overloads is useful 🙏

Goooler:
Seems we can return `Persistent*` as the value, WHYT?

hoc081098:
> Seems we can return `Persistent*` as the value, WHYT?

The change of return type to Persistent is inefficient.

For example, if the list is ImmutableList, but is not PersistentList, orEmpty will convert it to PersistentList, it is inefficient.

The better way is having two overloads, that avoids unnecessary operations.

Goooler:
I didn't take `immutable*Of` as the implementation due to the deprecations, they are not suggested to be used.

https://github.com/Kotlin/kotlinx.collections.immutable/blob/1d18389e32e8afdbf5caa89808fb5af308fb5c8d/core/commonMain/src/extensions.kt#L513-L570

qurbonzoda:
It makes sense to add the same extensions for `Persistent` counterparts as well.
However, let's postpone adding new extensions until we have a clear solution for handling the issue of stdlib overloads having precedence. See https://github.com/Kotlin/kotlinx.collections.immutable/issues/10 and https://github.com/Kotlin/kotlinx.collections.immutable/issues/181 for details.

Goooler:
CC @DmitryNekrasov

PR #214, opened 2025-03-27
Adopt KUP requirements imposed by KT-75078

### LINKED ISSUE #212: Adopt new requirements imposed on Kotlin user projects
See requirements here: https://youtrack.jetbrains.com/issue/KT-75078
We need to update build scripts to conform them.

fzhinkin:
`-Werror` doesn't work with `2.0.20`, so we either have to check the compiler version before using it, or update the Kotlin version (https://github.com/Kotlin/kotlinx.collections.immutable/pull/213).

Closes #212

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -63,8 +63,6 @@ allprojects {
                     "-Wextra",
                     "-Xuse-fir-experimental-checkers"
                 )
-                logger.info("kotlin_Werror_override was set to 'disable'")
-                logger.info("Using -Wextra and -Xuse-fir-experimental-checkers compiler options")
             }
             freeCompilerArgs.addAll(
                 "-Xexpect-actual-classes",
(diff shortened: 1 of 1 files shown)


woainikk:
Can you please also add logging for all of these flags? 
For example: 
 freeCompilerArgs.add("-Xrender-internal-diagnostic-names")
 buildScriptLogger.log(LogLevel.INFO, "-Xrender-internal-diagnostic-names enabled")
 freeCompilerArgs.add("-Wextra")
 buildScriptLogger.log(LogLevel.INFO, "-Wextra enabled")

woainikk:
And can you please also add this:
implement a Gradle property for passing additional CLI options to the Kotlin compiler
(property naming is not critical but **kotlin_additional_cli_options** is suggested for consistency)

ilya-g (approved):

ilya-g:
> -Werror doesn't work with 2.0.20

Did you mean `-Wextra`? 
Maybe we can assume that if `kotlin_Werror_override` is specified, the version of the compiler is also one that supports `-Wextra`?

build.gradle.kts
```diff
@@ -46,10 +46,31 @@ allprojects {
...
-            allWarningsAsErrors = true
-            freeCompilerArgs.add("-Xexpect-actual-classes")
+            if (setAllWarningsAsError.orNull != false) {
+                allWarningsAsErrors = true
+            } else {
+                freeCompilerArgs.addAll(
+                    "-Wextra",
+                    "-Xuse-fir-experimental-checkers"
+                )
+                logger.info("kotlin_Werror_override was set to 'disable'")
```
fzhinkin, line 66: @woainikk that'll emit a message for each and every compilation task, but I think it's better than writing it outside tasks configuration and them accidentally forgetting to setup `freeCompilerArgs`.

woainikk:
Checked on playground, everything works as expected!

woainikk (approved):

PR #219, opened 2025-05-07
Ghost element when applying multiples minus operations on a PersistentHashSet

DmitryNekrasov:
If after removing elements from one collision there is one element left, it should be promoted up recursively, but because the `realBitMap == bitmap` check is above the `realSize == 1 && shift != 0` check (and the first check always passes if the removal does not happen at this step of the recursion), the promotion only happens 1 level up. I swapped these checks, now everything works correctly. Also, I've added a simpler test that reproduces this issue.


fzhinkin (approved):
Nicely done!

PR #221, opened 2025-05-09
Release v0.4.0

DmitryNekrasov:
Add a badge with a Kotlin version.
If I understand correctly, the library version in the Readme will be updated automatically after publishing to maven central.

## 0.4.0

- Fixed the equality bug in PersistentMap — Added proper node promotion during mutable operations to ensure a consistent internal tree structure [#217](https://github.com/Kotlin/kotlinx.collections.immutable/pull/217)
- Fixed the bug in PersistentMap equals implementation — Ensured proper node promotion during mutable key removal even when nodes share the same owner [#218](https://github.com/Kotlin/kotlinx.collections.immutable/pull/218)
- Fixed the ghost element issue in PersistentHashSet — Corrected condition check order to ensure proper recursive element promotion after collision removal [#219](https://github.com/Kotlin/kotlinx.collections.immutable/pull/219)
- Updated Kotlin to version 2.1.20 and core dependencies [#213](https://github.com/Kotlin/kotlinx.collections.immutable/pull/213)
- Enabled '-Xjvm-default=disable' explicitly to prevent API dump changes [#210](https://github.com/Kotlin/kotlinx.collections.immutable/pull/210)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ kotlin {
     sourceSets {
         commonMain {
              dependencies {
-                 implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.3.8")
+                 implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.3.9")
              }
         }
     }
@@ -155,7 +155,7 @@ Add dependencies (you can also add other modules that you need):
 <dependency>
     <groupId>org.jetbrains.kotlinx</groupId>
     <artifactId>kotlinx-collections-immutable-jvm</artifactId>
-    <version>0.3.8</version>
+    <version>0.3.9</version>
 </dependency>
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,6 +5,8 @@
 [![GitHub license](https://img.shields.io/github/license/kotlin/kotlinx.collections.immutable)](LICENSE.txt) 
 [![Build status](https://teamcity.jetbrains.com/guestAuth/app/rest/builds/buildType:(id:KotlinTools_KotlinxCollectionsImmutable_Build_All)/statusIcon.svg)](https://teamcity.jetbrains.com/viewType.html?buildTypeId=KotlinTools_KotlinxCollectionsImmutable_Build_All)
 [![Maven Central](https://img.shields.io/maven-central/v/org.jetbrains.kotlinx/kotlinx-collections-immutable.svg?label=Maven%20Central)](https://central.sonatype.com/artifact/org.jetbrains.kotlinx/kotlinx-collections-immutable)
+[![Kotlin](https://img.shields.io/badge/kotlin-2.1.20-blue.svg?logo=kotlin)](http://kotlinlang.org)
```
fzhinkin, line 8: The min required Kotlin version is also explicitly mentioned somewhere below, it also needs to be updated.

fzhinkin:
Please also update the changelog file.

And I'm not sure if its worth mentioning all build and infra related changes (other than Kotlin version update) in the changelog and release notes - users don't care about it.

fzhinkin:
> If I understand correctly, the library version in the Readme will be updated automatically after publishing to maven central.

It won't. Only the version in the maven-central badge. Versions in the `Using in the projects`section needs to be updated manually.

fzhinkin (changes requested):

fzhinkin:
About the library version: Kotlin update may be considered as a significant change, that requires bumping up the version to `0.4.0` (but we didn't do that for previous Kotlin version updates for this library). @ilya-g WDYT?

README.md
```diff
@@ -137,7 +139,7 @@ kotlin {
     sourceSets {
         commonMain {
              dependencies {
-                 implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.3.8")
+                 implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.4")
```
fzhinkin, line 142: It should be `0.4.0`, not `0.4`.

fzhinkin (approved):

PR #227, opened 2025-06-18
Publish to Central

fzhinkin:
Setup publication to the maven central through an intermediate maven repo.


.teamcity/CentralDeployment.kt
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    steps {
+        gradle {
+            name = "Deploy ${platform.buildTypeName()} Binaries"
+            jdkHome = "%env.$jdk%"
+            jvmArgs = "-Xmx1g"
+            gradleParams =
+                "--info --stacktrace -P$versionSuffixParameter=WHAT -P$releaseVersionParameter=%$releaseVersionParameter%"
```
fzhinkin, line 129: TDB: I'm not 100% if suffix parameter works at all.

Issue #232, opened 2025-10-23
Rename PersistentCollection methods to participial forms and deprecate originals


DmitryNekrasov:
This issue implements a methodological refinement to the PersistentCollection interface to enhance API clarity and align with Kotlin standard library conventions.

## Problem

The current PersistentCollection interface uses imperative verb forms (add, remove, clear) that mirror MutableCollection's API, creating semantic ambiguity. While MutableCollection methods perform in-place mutations, PersistentCollection methods return new instances. This behavioral difference is not surfaced at the API level, requiring developers to rely on documentation rather than self-documenting code.

## Solution

Apply Kotlin's established naming pattern (like sort() vs sorted()) to distinguish mutating from non-mutating operations. The participial "-ing" suffix immediately communicates that operations produce new instances rather than altering existing state.

## Implementation

Introduce new methods with participial naming and deprecate existing imperative verb methods with migration guidance.

**PersistentCollection**: add → adding, addAll → addingAll, remove → removing, removeAll → removingAll, retainAll → retainingAll, clear → cleared

**PersistentList**: add(index) → addingAt(index), addAll(index) → addingAllAt(index), set(index) → replacingAt(index), removeAt → removingAt

**PersistentMap**: put → putting, putAll → puttingAll, remove(key) → removing(key), remove(key, value) → removing(key, value), clear → cleared

## What this looks like in practice

You can find usages of that new naming convention in real project (Thunderbird android app) [here](https://github.com/DmitryNekrasov/thunderbird-android/commit/1b53d63710673761ed98b2b3a0f6e565d51cfa47) or in the code snippets below:

```kotlin
private operator fun ImmutableList<String>.plus(other: String): ImmutableList<String> =
        (this.toMutableList() + other).toPersistentList()
```

```kotlin
private operator fun ImmutableList<String>.plus(other: String): ImmutableList<String> =
        toPersistentList().adding(other)
```

---

```kotlin
current.copy(bannerInlineVisuals = (current.bannerInlineVisuals + new).toPersistentSet())
```

```kotlin
current.copy(bannerInlineVisuals = current.bannerInlineVisuals.addingAll(new))
```

---

```kotlin
(current.bannerInlineVisuals - bannerInlineVisual).toPersistentSet()
```

```kotlin
current.bannerInlineVisuals.removing(bannerInlineVisual)
```

PR #233, opened 2025-10-27
Rename PersistentCollection methods to participial forms and deprecate originals, `*ing` | `*ed`

DmitryNekrasov:
#232


fzhinkin (changes requested):

core/commonMain/src/ImmutableCollection.kt
```diff
@@ -39,6 +53,20 @@ public interface PersistentCollection<out E> : ImmutableCollection<E> {
...
+     * @return a new persistent collection with elements of the specified [elements] collection added;
+     * or this instance if no modifications were made in the result of this operation.
+     *
+     * Imperative names mirror MutableCollection's mutating API.
+     * Participial form clarifies that a new instance is returned.
+     *
+     * @see addingAll
+     */
+    @Deprecated("Use addingAll() instead.", replaceWith = ReplaceWith("addingAll(elements)"))
     public fun addAll(elements: Collection<@UnsafeVariance E>): PersistentCollection<E>
```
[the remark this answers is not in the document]
DmitryNekrasov: Fixed

docs/0.5.0-MIGRATION.md
```diff
@@ -0,0 +1,103 @@
+# Migration Guide: Participial Naming for Copy-Returning Operations
```
fzhinkin, line 1: Assuming this might be not the one and only migration for the library, it makes sense to depict it in a name. Like `0.5.0-MIGRATION.md`.

Also, consider adding a `doc` directory to group various documents related to the library (like this one and `proposal.md`) there.

MIGRATION.md
```diff
@@ -0,0 +1,103 @@
...
+## Why this change
+
+Persistent collections are immutable, so every "mutating-looking" call actually
+returns a new collection. Using the same imperative names as `MutableCollection`
+(`list.add(x)`, `map.put(k, v)`, `list.removeAt(i)`) made this easy to misread at
+a call site: the method looks like it mutates `this`, but its result must be used
+or the call is effectively a no-op. This has been a recurring source of bugs and
+cognitive overhead during code review.
+
+KEEP-0459 establishes a uniform rule across Kotlin APIs: the imperative form is
```
fzhinkin: Small not: it makes sense to make all reference to the KEEP actual links.

docs/0.5.0-MIGRATION.md
```diff
@@ -0,0 +1,103 @@
...
+downstream projects have time to migrate without being forced into a flag-day
+upgrade.
+
+| Version | `DeprecationLevel` | What happens | Action required |
+| --- | --- | --- | --- |
+| `0.5.0` | `WARNING` — introduced in [#233](https://github.com/Kotlin/kotlinx.collections.immutable/pull/233) | Old names still compile. IDE shows deprecation warnings with `ReplaceWith` quick-fixes. Binary compatible. | None immediately. Start migrating at your own pace. |
+| `0.6.0` | `ERROR` | Old names still resolve at runtime (binary compatible) but no longer compile from source. | Migrate all call sites before upgrading. |
+| `0.7.0` | Removed | Old names are deleted from the API entirely. | Migration must be complete before upgrading. |
+
+## How to migrate
```
fzhinkin, line 89: Any recommendations to library authors who implement our interfaces?

MIGRATION.md
```diff
@@ -0,0 +1,103 @@
...
+| `0.7.0` | Removed | Old names are deleted from the API entirely. | Migration must be complete before upgrading. |
+
+## How to migrate
+
+- **Use the IDE quick-fix.** Every deprecated method carries
+  `@Deprecated(message, ReplaceWith(...))`. In IntelliJ IDEA / Android Studio,
+  place the caret on the warning and choose *Replace with …*, or run
+  *Code → Inspect Code → Deprecated API usages* to apply replacements across
+  the whole module.
+- **Bulk-migrate with the compiler.** On `0.5.x` you can temporarily raise the
```
fzhinkin: Well, it's not a migration, but a way to make all places requiring the migration more visible. I'm not sure it will be extremely helpful, especially if users can apply the previous suggestion and migration everything in IDE.

docs/0.5.0-MIGRATION.md
```diff
@@ -0,0 +1,103 @@
...
+  `@Deprecated(message, ReplaceWith(...))`. In IntelliJ IDEA / Android Studio,
+  place the caret on the warning and choose *Replace with …*, or run
+  *Code → Inspect Code → Deprecated API usages* to apply replacements across
+  the whole module.
+- **Bulk-migrate with the compiler.** On `0.5.x` you can temporarily raise the
+  deprecation warnings to errors (`-Werror` or per-file `@file:Suppress`
+  auditing) to make the remaining call sites easy to find.
+- **No behavior changes.** The new methods are exact renames — same semantics,
+  same return types, same complexity. No call-site logic needs to be adjusted.
+
```
fzhinkin, line 98: BTW, did you consider adding an AI agent skill that'll help with the migration? Assuming that many developers rely on it nowadays, it could be useful.

(I guess, adding a skill(s) dedicated to collections immutable in general, and to its migration in particular deserves a separate issue/PR, and the right place to host a skill(s) is https://github.com/Kotlin/kotlin-agent-skills).

docs/0.5.0-MIGRATION.md
DmitryNekrasov: Done.

MIGRATION.md
DmitryNekrasov: Done.

docs/0.5.0-MIGRATION.md
DmitryNekrasov: Done.

MIGRATION.md
DmitryNekrasov: Remove it.

docs/0.5.0-MIGRATION.md
DmitryNekrasov: It's really useful, but let's do it in a separate task.

core/commonMain/src/ImmutableList.kt
fzhinkin, line 20: Something we should decide before releasing the library: given the upcoming binary incompatible changes, that will affect implementors, should we annotate all public interface with `@SubclassOptInRequired`?

fzhinkin: (on the other hand, the annotation implies stability to use / instability to implement, but the upcoming changes will also affect users)

core/commonMain/src/extensions.kt
```diff
@@ -56,18 +56,19 @@ public inline fun <K, V> PersistentMap<out K, V>.mutate(mutator: (MutableMap<K,
 /**
  * Returns the result of adding the specified [element] to this collection.
  *
- * @returns a new persistent collection with the specified [element] added;
- * or this instance if this collection does not support duplicates and it already contains the element.
+ * @return a new persistent collection with the specified [element] added;
```
fzhinkin, line 59: Topic for a different PR: there's an inconsistency in how these operators are documented - this one has `@returns` with the description,  but operators overloaded for `PersistentList` don't.

core/commonMain/src/ImmutableCollection.kt
```diff
@@ -28,58 +28,163 @@ public interface PersistentCollection<out E> : ImmutableCollection<E> {
...
+     *         or this instance if this collection does not support duplicates,
+     *         and it already contains the element.
      */
+    public fun adding(element: @UnsafeVariance E): PersistentCollection<E> = @Suppress("DEPRECATION") add(element)
+
+    /**
+     * Returns the result of adding the specified [element] to this collection.
+     *
+     * Imperative names mirror MutableCollection's mutating API.
+     * Participial form clarifies that a new instance is returned.
```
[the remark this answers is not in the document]
fzhinkin: I agree with Ilya - the current description is indeed confusing unless you're deeply entrenched into specifics of the renaming. Just suggesting to use a different function should be enough.

To add more context and point users to the feedback channel for this PR, we can add (in a follow up PR, for example) a link to the migration guide with wording like this:
```
Use the function [adding] to make it clear that a new collection is return.

Old functions mimicking [MutableCollection] names, like this one, were deprecared and will be removed in future releases. Refer to the [Migration guide](https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/docs/0.5.0-MIGRATION.md) for more details and guidance with the migration.
```

And that, the deprecation message could be adjusted to:
```
@Deprecated("Use adding() instead. For more details, read the documentation for this function.", ...
```

fzhinkin (changes requested):
Besides the Ilya's comment about docs for deprecated functions, I have nothing else to add.

core/api/kotlinx-collections-immutable.api
```diff
@@ -106,30 +110,58 @@ public abstract interface class kotlinx/collections/immutable/ImmutableSet : jav
...
+	public abstract fun removingAll (Lkotlin/jvm/functions/Function1;)Lkotlinx/collections/immutable/PersistentCollection;
 	public abstract fun retainAll (Ljava/util/Collection;)Lkotlinx/collections/immutable/PersistentCollection;
+	public abstract fun retainingAll (Ljava/util/Collection;)Lkotlinx/collections/immutable/PersistentCollection;
 }
 
 public abstract interface class kotlinx/collections/immutable/PersistentCollection$Builder : java/util/Collection, kotlin/jvm/internal/markers/KMutableCollection {
 	public abstract fun build ()Lkotlinx/collections/immutable/PersistentCollection;
 }
 
+public final class kotlinx/collections/immutable/PersistentCollection$DefaultImpls {
```
ilya-g, line 132: Hmmm, I wonder why we still have DefaultImpls here? I thought we have migrated libraries to the other jvm-default modes. 

It's not to do something on this in this PR, but has to be sorted before the stabilization.

docs/0.5.0-MIGRATION.md
```diff
@@ -0,0 +1,132 @@
...
+    @Suppress("DEPRECATION")
+    override fun add(element: E): MyList<E> = adding(element)
+    ```
+
+  This avoids an extra virtual dispatch through the library's default
+  implementation, and the `0.7.0` migration is then just deleting the
+  deprecated override.
+- **At `0.6.0`**, the deprecation level becomes `ERROR`. Overrides of the
+  imperative methods need `@Suppress("DEPRECATION_ERROR")` rather than
```
ilya-g: Is it really so? I don't think it's a good practice to recommend suppressing errors, and it probably won't work without an additional opt-in with a compiler argument.

ilya-g (changes requested):
In general all ok, just a minor concern about migration guide.

docs/0.5.0-MIGRATION.md
DmitryNekrasov: Done.

core/commonMain/src/extensions.kt
DmitryNekrasov: https://github.com/Kotlin/kotlinx.collections.immutable/issues/254

docs/0.5.0-MIGRATION.md
DmitryNekrasov: https://github.com/Kotlin/kotlin-agent-skills/issues/43

ilya-g (approved):

fzhinkin (approved):

PR #247, opened 2026-02-26
Enable Kotlin Power Assert plugin

DmitryNekrasov:
This PR enables the Kotlin Power Assert compiler plugin, which transforms assertion calls to include detailed diagnostic messages showing intermediate expression values on failure.  
  
Power Assert injects a diagnostic message string as the last parameter of each configured assertion function. For `assertEquals(expected, actual)` this string is always constructed eagerly, even when the assertion passes. This has no impact on contract and unit tests, but stress tests execute assertions inside tight loops with up to 1000000 iterations, making the constant string allocation a significant performance bottleneck.  
  
To fix this, assertions inside loops in stress tests are replaced with `check` equivalents (`check(a == b)` instead of `assertEquals(a, b)`). Unlike `assertEquals`, `check` accepts a lambda for its message, so the diagnostic string is only materialized on failure. Assertions outside loops and in non-stress tests are left unchanged for readability. `kotlin.check` is already in the Power Assert functions list, so diagnostic output on failure is preserved.

### DIFF
--- a/core/commonTest/src/stress/list/PersistentListBuilderTest.kt
+++ b/core/commonTest/src/stress/list/PersistentListBuilderTest.kt
@@ -31,11 +31,11 @@ class PersistentListBuilderTest : ExecutionTimeMeasuringTest() {
         val elements = distinctStringValues(elementsToAdd)
         repeat(times = elementsToAdd) { index ->
             builder.add(elements[index])
-            check(!builder.isEmpty())
+            check(builder.isNotEmpty())
         }
         repeat(times = elementsToAdd - 1) {
             builder.removeAt(builder.size - 1)
-            check(!builder.isEmpty())
+            check(builder.isNotEmpty())
         }
         builder.removeAt(builder.size - 1)
         assertTrue(builder.isEmpty())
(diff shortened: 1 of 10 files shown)


core/commonTest/src/stress/list/PersistentListBuilderTest.kt
```diff
@@ -110,7 +110,7 @@ class PersistentListBuilderTest : ExecutionTimeMeasuringTest() {
         repeat(times = elementsToAdd) { index ->
             list.add(index)
             builder.add(index)
-            assertEquals<List<*>>(list, builder.toList())
+            check(list == builder.toList())
```
fzhinkin: `check` will throw `IAE`, Idea will not consider it a regular test failure and `<Click to see difference>` will be unavailable.

While it may not be an issue for assertion where two trivial values are compared, it will become an issue if two collections containing thousands of elements are not equal and you have to find the diff.

I would suggest waiting for a proper PA support from k-test and enable it then.

DmitryNekrasov:
Closing in favor of #270, which tracks doing this properly once the library is on Kotlin 2.4.20. The kotlin.check workaround here isn't worth keeping in the meantime.

PR #251, opened 2026-04-24
Fix ordered set builder cache after removeAll

odenix:
This bug was found by Codex by adding randomized trace tests that exercised persistent collections and builders through adversarial operation sequences, checked each step against JDK reference collections, and verified snapshot immutability across rebuilds. The fix in this change was also produced by Codex.

A recursive mutable remove in immutableMap/TrieNode could rebuild a parent node even when the child remove was a logical no-op. That invalidated PersistentHashMapBuilder.builtMap without any content change.

PersistentOrderedSetBuilder caches a built set and asserts that its backing hashMapBuilder still has a builtMap. After certain remove/removeAll and build sequences, that assertion could fail because the hash map builder cache had been cleared spuriously.

Fix mutableReplaceNode to preserve node identity for true no-op child removes, while still allowing parent updates when the child becomes promotable to a direct entry.

Add a deterministic regression test in PersistentOrderedSetTest covering the repeated removal/rebuild sequence that previously tripped the builder cache assertion.


core/commonTest/src/contract/set/PersistentOrderedSetBuilderTest.kt
```diff
@@ -0,0 +1,139 @@
...
+
+        fun builderRemoveAll(vararg values: Int) {
+            val keys = keys(*values).toSet()
+            builder.removeAll(keys)
+            expectedBuilder.removeAll(keys)
+        }
+
+        fun persistentAdd(value: Int) {
+            val key = key(value)
+            persistent = persistent.add(key)
```
DmitryNekrasov: Could you rebase onto current master and rename the persistent-collection calls? They were renamed per KEEP-0459 after you opened this, so once rebased the test no longer compiles. Only the `persistent.*` calls change: `add`→`adding`, `addAll`→`addingAll`, `remove`→`removing`, `removeAll`→`removingAll`. The builder calls stay as they are.

core/commonTest/src/contract/set/PersistentOrderedSetBuilderTest.kt
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+package tests.contract.set
+
+import kotlinx.collections.immutable.persistentSetOf
+import kotlin.collections.LinkedHashSet
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class PersistentOrderedSetBuilderTest {
```
DmitryNekrasov, line 13: `PersistentOrderedMapBuilder` has the same bug, but your change fixes it too. Could you please add an analogous regression test for the map builder?

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -763,16 +763,18 @@ internal class TrieNode<K, V>(
...
 
         // key is absent
         return this
     }
 
-    private fun mutableReplaceNode(newNode: TrieNode<K, V>?, nodeIndex: Int, positionMask: Int, owner: MutabilityOwnership) = when {
+    private fun mutableReplaceNode(targetNode: TrieNode<K, V>, newNode: TrieNode<K, V>?, nodeIndex: Int, positionMask: Int, owner: MutabilityOwnership) = when {
         newNode == null ->
             mutableRemoveNodeAtIndex(nodeIndex, positionMask, owner)
+        newNode === targetNode && !(newNode.buffer.size == 2 && newNode.nodeMap == 0) ->
```
DmitryNekrasov: Let's add a comment to explain this behavior.

```kotlin
// newNode === targetNode means the child returned itself (a no-op, or an owned in-place removal),
// so this node's buffer already points to it. Keep this node unchanged to avoid spuriously
// clearing `PersistentHashMapBuilder.builtMap` on no-ops. The single-entry exclusion still routes a
// child that shrank to one entry to `updateNodeAtIndex`, which promotes it.
newNode === targetNode && !(newNode.buffer.size == 2 && newNode.nodeMap == 0) -> this
```

DmitryNekrasov (dismissed):
The fix is right. It brings the mutable remove path in line with the immutable `replaceNode`. Before merge, see comments: it doesn't compile on current master, and `PersistentOrderedMapBuilder` has the same bug, untested.

DmitryNekrasov:
Hey @odenix! Are you going to work on this PR, or can I make the appropriate changes? Thank you!

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -763,16 +763,26 @@ internal class TrieNode<K, V>(
...
+        nodeIndex: Int,
+        positionMask: Int,
+        owner: MutabilityOwnership
+    ) = when {
+        newNode == null -> mutableRemoveNodeAtIndex(nodeIndex, positionMask, owner)
+        // `newNode` === `targetNode` means the child returned itself (a no-op, or an owned in-place removal),
+        // so this node's buffer already points to it. Keep this node unchanged to avoid spuriously
+        // clearing `PersistentHashMapBuilder.builtMap` on no-ops. The single-entry exclusion still routes a
+        // child that shrank to one entry to `updateNodeAtIndex`, which promotes it.
+        newNode === targetNode && !(newNode.buffer.size == 2 && newNode.nodeMap == 0) -> this
```
ilya-g: Suggestion: I see that this condition is checked in multiple places, so it may make sense to extract it in a more meaningfully named property 
`newNode.buffer.size == 2 && newNode.nodeMap == 0`

ilya-g (dismissed):

core/commonMain/src/implementations/immutableMap/TrieNode.kt
DmitryNekrasov: Done. Added `hasSingleEntry` property.

ilya-g (approved):

PR #252, opened 2026-04-25
Fix ordered map builder entry value updates

odenix:
Fixes a stale cached-map result when updating a PersistentMap builder entry through MutableMap.MutableEntry.setValue.

PersistentOrderedMapBuilder caches the last built ordered map. Entry setValue previously updated only the underlying hash-map builder, leaving the ordered builder cache valid. A subsequent build() could therefore return the previous map.

This change routes entry value updates through the ordered-map builder so the ordered builder cache is invalidated before the hash-map builder is updated.

Added a regression test that builds once, updates an entry value through the builder entry iterator, and verifies the next build() reflects the new value.


odenix:
Closing for now; this was opened prematurely while preparing the fix branch.

PR #253, opened 2026-04-25
Invalidate ordered map builder after entry value updates

odenix:
This bug was found and fixed by Codex.

PersistentOrderedMapBuilder caches the last built ordered map. Updating an entry through MutableMap.MutableEntry.setValue updated only the backing hash map builder, leaving the ordered builder cache incorrectly valid. A later build() could therefore return the previously built map.

Route entry value updates through the ordered map builder so its existing cache invalidation path runs. Add a regression test that builds once before mutating an entry value.


DmitryNekrasov (approved):
Fix is correct.

PR #255, opened 2026-05-12
Add top-level README for kotlinx-collections-immutable module

DmitryNekrasov:
## Summary
- Adds `core/README.md` with module-level documentation: motivation, immutable vs persistent semantics, quick-start snippets for `persistentListOf` / `persistentSetOf` / `persistentMapOf`, `+`/`-` operators, `mutate { }` for batched edits, and the list of supported Kotlin Multiplatform targets.
- Wires the README into Dokka via `includes.from("README.md")` in `core/build.gradle.kts` so it renders as the module overview, together with the `# Package kotlinx.collections.immutable` and `# Package kotlinx.collections.immutable.adapters` sections from the same file.

## Test plan
- [x] Run the Dokka task locally and confirm the generated module page shows the README content.
- [x] Confirm per-package descriptions for `kotlinx.collections.immutable` and `kotlinx.collections.immutable.adapters` appear on their package pages.
- [x] Eyeball the rendered Markdown for broken links or code-block formatting issues.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -1,0 +1,118 @@
+# Module kotlinx-collections-immutable
+
+Immutable and persistent collection types for Kotlin Multiplatform.
+
+This library provides interfaces and efficient implementations of read-only and persistent
+variants of lists, sets, and maps. It complements the Kotlin standard library, where
+collection interfaces describe the absence of mutating methods on the caller but do not
+guarantee that the underlying instance cannot be mutated elsewhere.
+
+## Why immutable collections
+
+Most Kotlin code relies on `List`, `Set`, and `Map` — interfaces that omit mutating methods
+but allow the underlying implementation to be mutable. Passing such a collection across
+module boundaries does not protect it from being modified by the caller or by another
+holder of the same reference.
+
+Immutable collections close this gap:
+
+- **Safe sharing.** Once constructed, the value cannot change, so it can be freely shared
+  across threads or kept as part of a state snapshot without defensive copies.
+- **Predictable equality.** Hash codes are stable, which makes immutable collections safe
+  to use as map keys, set elements, and cache values.
+- **Stable in Compose.** Jetpack Compose recognizes persistent collections as stable, so a
+  composable that takes one as a parameter can skip recomposition when its contents are
+  unchanged.
+
+## Immutable vs persistent
+
+The library defines two interface families:
+
+- **Immutable** — `ImmutableList`, `ImmutableSet`, `ImmutableMap`. These extend the
+  standard read-only collection interfaces and add a contract that the underlying value
+  cannot change after construction. They expose no operations to derive a modified copy.
+- **Persistent** — `PersistentList`, `PersistentSet`, `PersistentMap`. These extend the
+  immutable interfaces and add copy-returning operations such as `adding`, `removing`,
+  and `putting`. The returned collection shares structure with the original, so most
+  modifications run in near-constant or logarithmic time and avoid copying the entire
+  collection.
+
+Use persistent variants whenever a collection needs to evolve — to keep an event log,
+accumulate intermediate results, or maintain immutable state in a reducer or UI store.
+Use plain immutable variants for fully baked values that will not change.
+
+## Quick start
+
+```kotlin
+import kotlinx.collections.immutable.*
+
+val numbers = persistentListOf(1, 2, 3)
+val extended = numbers.adding(4)                    // PersistentList: [1, 2, 3, 4]
+
+val tags = persistentSetOf("kotlin", "immutable")
+val withMore = tags.adding("multiplatform")         // PersistentSet: [kotlin, immutable, multiplatform]
+
+val config = persistentMapOf("retries" to 3, "timeout" to 30)
+val updated = config.putting("timeout", 60)         // PersistentMap: {retries=3, timeout=60}
+
+val snapshot: PersistentList<Int> = listOf(1, 2, 3).toPersistentList()
+```
+
+The default `persistentSetOf` and `persistentMapOf` preserve element insertion order
+during iteration. If iteration order does not matter, the unordered implementations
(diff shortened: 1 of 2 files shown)


core/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+The builder seen inside `mutate` implements the standard `MutableList`, `MutableSet`,
+or `MutableMap` interface, so any code that already works with mutable collections can
+be reused inside the block.
+
+## Multiplatform support
+
+`kotlinx-collections-immutable` is a Kotlin Multiplatform library. It is published for
+JVM, JS, Wasm (JS and WASI), and all Kotlin/Native
+[targets](https://kotlinlang.org/docs/native-target-support.html) supported by the
```
fzhinkin: Maybe we can say "all targets are supported" instead? And then give a list, if it is needed.

core/README.md
```diff
@@ -0,0 +1,118 @@
...
+  to use as map keys, set elements, and cache values.
+- **Stable in Compose.** Jetpack Compose recognizes persistent collections as stable, so a
+  composable that takes one as a parameter can skip recomposition when its contents are
+  unchanged.
+
+## Immutable vs persistent
+
+The library defines two interface families:
+
+- **Immutable** — `ImmutableList`, `ImmutableSet`, `ImmutableMap`. These extend the
```
fzhinkin: Let's make all references to classes, interfaces and functions actual links (the same applies to links to stdilb interfaces, like Map or List):
```
[`ImmutableList`][kotlinx.collections.immutable.ImmutableList]
```
or
```
[ImmutableList][kotlinx.collections.immutable.ImmutableList]
```

core/README.md
```diff
@@ -0,0 +1,118 @@
...
+variants of lists, sets, and maps. It complements the Kotlin standard library, where
+collection interfaces describe the absence of mutating methods on the caller but do not
+guarantee that the underlying instance cannot be mutated elsewhere.
+
+## Why immutable collections
+
+Most Kotlin code relies on `List`, `Set`, and `Map` — interfaces that omit mutating methods
+but allow the underlying implementation to be mutable. Passing such a collection across
+module boundaries does not protect it from being modified by the caller or by another
+holder of the same reference.
```
fzhinkin: I see the point and fully agree with it, but I believe we can slightly tune the wording to clarify what stdlib collections are actually guarantee and what their shortcomings are.

For example, if a function accepts a `List`, but the caller has a reference to a `MutableList` and it passes it to the function, then yes, the function's contract is "I won't modify the list", but the caller still can modify it, so the function can't assume that the list will remain unchanged for eternity and cache it.

At the same time, if you got a `List` reference via `listOf(...)` call, then the list will not be "normally" modified and you can rely on it (to some extent).

core/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+Immutable collections close this gap:
+
+- **Safe sharing.** Once constructed, the value cannot change, so it can be freely shared
+  across threads or kept as part of a state snapshot without defensive copies.
+- **Predictable equality.** Hash codes are stable, which makes immutable collections safe
+  to use as map keys, set elements, and cache values.
+- **Stable in Compose.** Jetpack Compose recognizes persistent collections as stable, so a
```
fzhinkin: I don't think we should refer to some particular consumers of the library in its docs. But if we agree to keep it, it should be "Compose Multiplatform" (with a link to its documentation).

fzhinkin: I mean, the library is good not because CMP has some optimization that relies on it, but it is good and has some qualities that allowed CMP to apply some optimizations.

core/README.md
```diff
@@ -0,0 +1,118 @@
...
+  standard read-only collection interfaces and add a contract that the underlying value
+  cannot change after construction. They expose no operations to derive a modified copy.
+- **Persistent** — `PersistentList`, `PersistentSet`, `PersistentMap`. These extend the
+  immutable interfaces and add copy-returning operations such as `adding`, `removing`,
+  and `putting`. The returned collection shares structure with the original, so most
+  modifications run in near-constant or logarithmic time and avoid copying the entire
+  collection.
+
+Use persistent variants whenever a collection needs to evolve — to keep an event log,
+accumulate intermediate results, or maintain immutable state in a reducer or UI store.
```
fzhinkin: A map-reduce's reducer or what do you mean?

core/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Multiplatform support
+
+`kotlinx-collections-immutable` is a Kotlin Multiplatform library. It is published for
+JVM, JS, Wasm (JS and WASI), and all Kotlin/Native
+[targets](https://kotlinlang.org/docs/native-target-support.html) supported by the
+compiler.
+
+# Package kotlinx.collections.immutable
+
```
fzhinkin, line 131: There is also `kotlinx.collections.immutable.implementations.immutableList` package and it will be displayed on the rendered module-level page, so it's worth writing a few words about it.

fzhinkin (changes requested):
Nice document! There are a few comments from me here and there, but in general it already looks pretty good.

core/README.md
DmitryNekrasov: Done.

core/README.md
DmitryNekrasov: Done.

core/README.md
DmitryNekrasov: Done.

core/README.md
DmitryNekrasov: Done.

core/README.md
DmitryNekrasov: I meant a Redux- or MVI-style state reducer — the pure `(state, action) -> newState` function used by state containers. Reworded the line to spell out the intent.

core/README.md
DmitryNekrasov: Done.

fzhinkin (approved):

PR #256, opened 2026-05-13
Document return values in the KDoc summary instead of `@return`

### LINKED ISSUE #254: Unify `@return` KDoc style library-wide
While reviewing #233, @fzhinkin noted that the operator overloads are documented inconsistently — some carry a `@return` tag, while the `PersistentList` overloads describe the return in the summary instead ([discussion](https://github.com/Kotlin/kotlinx.collections.immutable/pull/233#discussion_r3169086789)). The same split runs through the rest of the public API, so the `@return` style should be unified library-wide.

Kotlin's [coding conventions](https://kotlinlang.org/docs/coding-conventions.html#documentation-comments) settle the direction: avoid `@return`, put the description in the main text, and use the tag only for a long description that doesn't fit there.

So the style is no `@return` — the return value goes into the summary, e.g. `Returns a new persistent list with the specified [element] appended`, including any "or this instance if …" case.

DmitryNekrasov:
Drops `@return` across the project, moving the return value into the KDoc summary.

`@param` is kept in the two cases the convention still allows: type parameters (documented as in the stdlib), and the internal `PersistentVector` constructor, whose parameter descriptions are too long to inline.

Fixes #254

### DIFF
--- a/core/commonMain/src/ImmutableList.kt
+++ b/core/commonMain/src/ImmutableList.kt
@@ -20,9 +20,11 @@ import kotlinx.collections.immutable.internal.ListImplementation
 public interface ImmutableList<out E> : List<E>, ImmutableCollection<E> {
 
     /**
-     * Returns a view of the portion of this list between the specified [fromIndex] (inclusive) and [toIndex] (exclusive).
+     * Creates a view of the portion of this list between the specified [fromIndex] (inclusive) and [toIndex] (exclusive).
      *
      * The returned list is backed by this list.
+     *
+     * @return a view of the portion of this list between the specified [fromIndex] (inclusive) and [toIndex] (exclusive).
      *
      * @throws IndexOutOfBoundsException if [fromIndex] is less than zero or [toIndex] is greater than the size of this list.
      * @throws IllegalArgumentException if [fromIndex] is greater than [toIndex].
@@ -65,12 +67,14 @@ public interface ImmutableList<out E> : List<E>, ImmutableCollection<E> {
  */
 public interface PersistentList<out E> : ImmutableList<E>, PersistentCollection<E> {
     /**
-     * Returns a new persistent list with the specified [element] appended.
+     * Returns the result of appending the specified [element] to this list.
+     *
+     * @return a new persistent list with the specified [element] appended.
      */
     override fun adding(element: @UnsafeVariance E): PersistentList<E> = @Suppress("DEPRECATION") add(element)
 
     /**
-     * Returns a new persistent list with the specified [element] appended.
+     * Returns the result of appending the specified [element] to this list.
      *
      * Use the function [adding] to make it clear that a new list is returned.
      *
@@ -78,6 +82,8 @@ public interface PersistentList<out E> : ImmutableList<E>, PersistentCollection<
      * were deprecated and will be removed in future releases. Refer to the
      * [Migration guide](https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/docs/0.5.0-MIGRATION.md)
      * for more details and guidance with the migration.
+     *
+     * @return a new persistent list with the specified [element] appended.
      */
     @Deprecated(
         "Use adding() instead. For more details, read the documentation for this function.",
@@ -237,12 +243,14 @@ public interface PersistentList<out E> : ImmutableList<E>, PersistentCollection<
     override fun retainAll(elements: Collection<@UnsafeVariance E>): PersistentList<E>
 
     /**
-     * Returns an empty persistent list.
+     * Returns the result of removing all elements from this list.
+     *
+     * @return an empty persistent list.
      */
     override fun cleared(): PersistentList<E> = @Suppress("DEPRECATION") clear()
 
     /**
-     * Returns an empty persistent list.
+     * Returns the result of removing all elements from this list.
      *
      * Use the function [cleared] to make it clear that a new list is returned.
      *
@@ -250,6 +258,8 @@ public interface PersistentList<out E> : ImmutableList<E>, PersistentCollection<
      * were deprecated and will be removed in future releases. Refer to the
      * [Migration guide](https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/docs/0.5.0-MIGRATION.md)
      * for more details and guidance with the migration.
+     *
+     * @return an empty persistent list.
      */
     @Deprecated(
         "Use cleared() instead. For more details, read the documentation for this function.",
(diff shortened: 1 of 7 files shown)


core/commonMain/src/extensions.kt
```diff
@@ -514,332 +516,412 @@ public operator fun <K, V> PersistentMap<out K, V>.minus(keys: Array<out K>): Pe
...
+ *
+ * @return a new persistent map with the specified contents.
  */
 @Deprecated("Use persistentHashMapOf instead.", ReplaceWith("persistentHashMapOf(*pairs)"))
 public fun <K, V> immutableHashMapOf(vararg pairs: Pair<K, V>): PersistentMap<K, V> = persistentHashMapOf(*pairs)
 
 
 /**
- * Returns an immutable list containing all elements of this collection.
+ * Converts this collection to an immutable list.
```
fzhinkin: It's rather iterable, not a collection

core/commonMain/src/ImmutableList.kt
```diff
@@ -65,19 +67,23 @@ public interface ImmutableList<out E> : List<E>, ImmutableCollection<E> {
  */
 public interface PersistentList<out E> : ImmutableList<E>, PersistentCollection<E> {
     /**
-     * Returns a new persistent list with the specified [element] appended.
+     * Returns the result of appending the specified [element] to this list.
```
fzhinkin: Could you elaborate why do you think "Returns the result of ..." version is better than what was written before?

Previous version was explicit about the new list being returned, and the new version gives an impression that the list will be modified.

Also, "Returns the result of ..." looks excessive, it makes sense to describe an operation in terms of actions and it should be obvious that it returns the result of that action.

core/commonMain/src/extensions.kt
```diff
@@ -514,332 +516,412 @@ public operator fun <K, V> PersistentMap<out K, V>.minus(keys: Array<out K>): Pe
...
+ *
+ * @return a new persistent list of the specified elements.
  */
 public fun <E> persistentListOf(vararg elements: E): PersistentList<E> = persistentVectorOf<E>().addingAll(elements.asList())
 
 /**
- * Returns an empty persistent list.
+ * Creates an empty persistent list.
+ *
+ * @return an empty persistent list.
```
JakeWharton: Lots of these are just noise. I don't think many of these changes offer an actual improvement.

https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/

core/commonMain/src/extensions.kt
DmitryNekrasov: Fixed, here and on the other `to*` functions with an `Iterable` receiver.

core/commonMain/src/ImmutableList.kt
DmitryNekrasov: Dropped `Returns the result of …` — it did read like the receiver gets mutated. The summary now names the new value directly: `Returns a new persistent list with the specified [element] appended`, no `@return` tag.

qwwdfsad:
The comment section duplication seems really redundant (and noisy) and harder to maintain in the long run.

I'd suggest fixing the KDoc spec/IDEA&Dokka rendering instead if there is a real need to have `@return` tag like that, not making the problem with the spec more severe

DmitryNekrasov:
@qwwdfsad the duplication is gone. I dropped `@return` and moved the return value into the summary, which is what the Kotlin [coding conventions](https://kotlinlang.org/docs/coding-conventions.html#documentation-comments) recommend, so there's no real need for the tag.

JakeWharton:
So as far as the new set of changes goes, it is more in line with what I prefer.

I will decline to submit a review (not that mine matters anyway), however, since you are clearly just a proxy for an LLM and not providing meaningful value as a human. I do not want to speak to an LLM. If I did, I could do so directly. Responding to what appears to be your coworkers with LLM slop is such an offensive thing to me. This pull request is a deeply troubling interaction, and I hope it is not indicative of what we can expect in the future for this project and others as first-party Kotlin offerings.

If for some reason you are using an LLM to overcome a language barrier, please consider using a model solely focused on translation and not one which steals your voice and agency.

DmitryNekrasov:
@JakeWharton a wall of agent-generated text is just noise for a reviewer. That was a mistake here, and I've shortened the description and the comment threads.

The change is mechanical, so I use an agent for the work, but I read and approve everything before it's posted, and the decisions are mine. What I missed was keeping the output short — that's on me, and it's what I'm fixing. We've discussed the same thing on the team: using AI is fine, but what we send to reviewers should be short and easy to check.

core/commonMain/src/extensions.kt
```diff
@@ -139,245 +113,192 @@ public operator fun <E> PersistentCollection<E>.minus(elements: Sequence<E>): Pe
...
         = if (elements is Collection) removingAll(elements) else mutate { it.removeAll(elements) }
 
 /**
- * Returns the result of removing all elements in this list that are also
- * contained in the specified [elements] array.
- *
- * @return a new persistent list with elements in this list that are also
- * contained in the specified [elements] array removed;
+ * Returns a new persistent list with elements in this list that are also
+ * contained in the specified [elements] array removed,
```
fzhinkin: If the kdoc is about to change anyway, let's reuse the way this operation is described for the stdlib's [minus](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/minus.html) (well, I don't reuse it literally, but rather inspired by it):
```
Returns  a new persistent list containing all elements of this list except the elements contained in the given [elements] array,
or this instance if there are no elements to remove.
```
Or something similar.

The current description is not that easy to follow, although it sheds the light on a pretty simple operation.

core/commonMain/src/extensions.kt
```diff
@@ -139,245 +113,192 @@ public operator fun <E> PersistentCollection<E>.minus(elements: Sequence<E>): Pe
...
  * or this instance if no modifications were made in the result of this operation.
```
fzhinkin: Also, as a topic for a separate discussion - should we really stress referential identity of the returned list? Does anyone cares that it will be new or the same instance? Perhaps, it should only be an implementation note, not the part of the contract.

fzhinkin: I guess, I have the problem with "if no modifications were made" part because the operation does not modify the receiver anyway. :/

core/commonMain/src/extensions.kt
DmitryNekrasov: Done.

core/commonMain/src/extensions.kt
DmitryNekrasov: Done.

fzhinkin (approved):

PR #258, opened 2026-05-13
Fix the package-list location

whyoleg:
This effectively makes the output look like a multi-module build, with `package-list` in the correct location.
See https://github.com/Kotlin/kotlinx-datetime/pull/580 / https://github.com/Kotlin/kotlinx-datetime/issues/268


DmitryNekrasov (approved):
LGTM

PR #259, opened 2026-05-13
Add `.kotlin` to `.gitignore` to exclude build metadata produced by KGP

whyoleg:
See https://kotlinlang.org/docs/gradle-configure-project.html#kotlin-gradle-plugin-data-in-a-project

Otherwise, you will see something like:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.iml
 target
 build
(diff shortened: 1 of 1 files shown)


DmitryNekrasov (approved):
LGTM

PR #268, opened 2026-06-22
Provide a JPMS module descriptor for the JVM artifact

### LINKED ISSUE #157: Provide module-info descriptor for java 9 and later

DmitryNekrasov:
Implements the module descriptor agreed in #157, minimal as requested on #175: compiled in the `core` build, no `buildSrc`.

`module-info.java` is compiled with `--release 9` into `META-INF/versions/9`, so the main classes stay Java 8-compatible and only Java 9+ sees the module `kotlinx.collections.immutable`. It exports the three packages with public API.

Modular consumer that runs on the module path: https://github.com/DmitryNekrasov/kotlinx-immutable-jpms-demo

Fixes #157


core/build.gradle.kts
```diff
@@ -193,3 +193,30 @@ tasks {
...
+    val modulePath = jvmMainCompilation.compileDependencyFiles
+
+    dependsOn(jvmMainCompilation.compileTaskProvider)
+    source(layout.projectDirectory.dir("jvmMain/java9"))
+    destinationDirectory.set(layout.buildDirectory.dir("classes/module-info/jvm"))
+    classpath = files()
+
+    javaCompiler.set(javaToolchains.compilerFor { languageVersion.set(JavaLanguageVersion.of(21)) })
+    options.release.set(9)
+    options.compilerArgs.add("-Xlint:-requires-transitive-automatic")
```
fzhinkin: Why this flag is needed?

core/build.gradle.kts
```diff
@@ -193,3 +193,30 @@ tasks {
...
+    val jvmMainCompilation = kotlin.jvm().compilations.getByName("main")
+    val kotlinClasses = jvmMainCompilation.output.classesDirs
+    val modulePath = jvmMainCompilation.compileDependencyFiles
+
+    dependsOn(jvmMainCompilation.compileTaskProvider)
+    source(layout.projectDirectory.dir("jvmMain/java9"))
+    destinationDirectory.set(layout.buildDirectory.dir("classes/module-info/jvm"))
+    classpath = files()
+
+    javaCompiler.set(javaToolchains.compilerFor { languageVersion.set(JavaLanguageVersion.of(21)) })
```
fzhinkin: Perhaps, it's time to factor out the toolchain version into a dedicated variable (since it is also used somewhere in a buildscript above)?

core/build.gradle.kts
```diff
@@ -193,3 +193,30 @@ tasks {
...
+    options.compilerArgs.add("-Xlint:-requires-transitive-automatic")
+    options.compilerArgumentProviders.add(CommandLineArgumentProvider {
+        listOf(
+            "--module-path", modulePath.asPath,
+            "--patch-module", "$moduleName=${kotlinClasses.asPath}",
+        )
+    })
+}
+
+tasks.named<Jar>("jvmJar") {
```
fzhinkin: Generated jar file includes `META-INF/versions/9/previous-compilation-data.bin` - is it meant to be like that?

core/build.gradle.kts
```diff
@@ -193,3 +193,30 @@ tasks {
...
+
+    javaCompiler.set(javaToolchains.compilerFor { languageVersion.set(JavaLanguageVersion.of(21)) })
+    options.release.set(9)
+    options.compilerArgs.add("-Xlint:-requires-transitive-automatic")
+    options.compilerArgumentProviders.add(CommandLineArgumentProvider {
+        listOf(
+            "--module-path", modulePath.asPath,
+            "--patch-module", "$moduleName=${kotlinClasses.asPath}",
+        )
+    })
```
fzhinkin: Should we also specify the module version? Akin to https://github.com/Kotlin/kotlinx-datetime/pull/593/changes

core/build.gradle.kts
```diff
@@ -193,3 +193,30 @@ tasks {
...
+}
+
+tasks.named<Jar>("jvmJar") {
+    from(compileJvmModuleInfo) { into("META-INF/versions/9") }
```
fzhinkin: Also, wouldn't it be easier to put the compiled module info directly into original source set's destination directory by specified a corresponding destinationDirectory?

core/jvmMain/java9/module-info.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2016-2026 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+module kotlinx.collections.immutable {
+    requires transitive kotlin.stdlib;
+
+    exports kotlinx.collections.immutable;
+    exports kotlinx.collections.immutable.adapters;
+    exports kotlinx.collections.immutable.implementations.immutableList;
```
fzhinkin, line 11: If other classes from implementations packages (like immutableMap) will become public, is there a way to catch a missing `exports` here?

core/build.gradle.kts
DmitryNekrasov: Done, extracted `jdkToolchainVersion`.

core/build.gradle.kts
DmitryNekrasov: Done.

core/build.gradle.kts
DmitryNekrasov: Done. It now follows datetime's setup.

core/build.gradle.kts
DmitryNekrasov: Silences javac's requires-transitive-automatic warning. Added comment, by analogue with datetime. For datetime it says that this flag `not needed when compiling with recent JDKs, e.g. 17`, but I see the warning for 17 and 21 as well.

core/build.gradle.kts
DmitryNekrasov: Not intended — it's Gradle's incremental-compilation file, copied because the jar took the whole output dir. Fixed: now only `module-info.class` is copied into `META-INF/versions/9/`.

core/jvmMain/java9/module-info.java
DmitryNekrasov: Not at compile time. The practical tripwire is the binary-compatibility `.api` dump — a newly public type in an `implementations.*` package shows up there, which is the cue to add the matching `exports`. I can try to add the test for that.

DmitryNekrasov: Added `ModuleInfoExportsTest`.

core/build.gradle.kts
fzhinkin: But how do you get an error? I get it when switching the toolchain to JDK 11, but with 21 there are no warnings (that's where my initial question is coming from - it seems like the flag is redundant).

core/jvmMain/java9/module-info.java
fzhinkin: I like the idea of having an automated test, but I doubt it should be a part of the regular unit test suite. Mostly because it verifies what would become a part of the final artifact, rather than some actual functionality. And it implicitly depends on `updateLegacyAbi` task's output.

The latter is an actual issue: Gradle is very upset when you try to run both `jvmTest` and `updateLegacyAbi` simultaneously.

It might be better to just register a new Gradle task that will perform the check and use `abiValidation.legacyDump.legacyDumpTaskProvider`'s `dumpDir` as an input (or `referenceDumpDir` + a dependency on `legacyUpdateTaskProvider`, depending on how _aggressive_ the check should be). CC @shanshin

core/build.gradle.kts
fzhinkin: I meant that you can actually set `compileJvmModuleInfo`'s destination directory to `compileKotlinJvm.destinationDirectory.dir("META-INF/versions/9")`, but it seems like `jvmJar` will still require a dependency on `compileJvmModuleInfo` to make it work. 
But yeah, it looks ok now.

core/build.gradle.kts
DmitryNekrasov: I got warnings on 17/21 in my demo app with a synthetic automatic module (w/o module-info). In the real build no warning occurs, so the flag is a no-op, removing it.

core/jvmMain/java9/module-info.java
DmitryNekrasov: Done. Replaced the test with a `checkModuleInfoExports` task that takes `legacyDump.legacyDumpTaskProvider`'s `dumpDir` as an input and is wired into `check`.

core/build.gradle.kts
```diff
@@ -198,10 +201,106 @@ tasks {
...
+        description = "Checks that jvmMain/java9/module-info.java exports exactly the public-API packages."
+
+        @OptIn(ExperimentalAbiValidation::class)
+        val dumpDir = kotlin.abiValidation.legacyDump.legacyDumpTaskProvider.flatMap { it.dumpDir }
+        val moduleInfoFile = layout.projectDirectory.file("jvmMain/java9/module-info.java")
+
+        inputs.file(moduleInfoFile).withPropertyName("moduleInfo")
+        inputs.dir(dumpDir).withPropertyName("legacyDumpDir")
+
+        val exportsRegex = Regex("""exports\s+([\w.]+)\s*;""")
```
fzhinkin, line 215: Just a fun fact - one can surround an export with block comments (`/* */`) and the check won't fail. I don't think there is anything we should do about it beside abstaining from commenting out lines in module-info.java :)

core/build.gradle.kts
```diff
@@ -198,10 +201,106 @@ tasks {
...
+                    Add `exports <package>;` for each.
+                    """.trimIndent()
+                )
+                if (stale.isNotEmpty()) sb.appendLine(
+                    """
+                    Packages exported by jvmMain/java9/module-info.java with NO public API in the ABI dump: $stale.
+                    Remove each stale `exports`.
+                    """.trimIndent()
+                )
+                throw GradleException(sb.toString().trim())
```
fzhinkin: Just curious, what are you trimming here? I mean, the string is created a few lines above, and we have full control over its content - why don't we build it the way no trimming is necessary?

fzhinkin (dismissed):

core/build.gradle.kts
DmitryNekrasov: It's just to avoid a trailing `\n` at the end. Rewritten using `buildString` and prefix `appendLine()`.

PR #274, opened 2026-07-02
Restore the direct links write in the ordered map builder entry

DmitryNekrasov:
Routing entry value updates through the builder's put fixed the stale build() result but made setValue look up the links the entry already caches. Now the entry writes them back directly while the hash builder's mod count is unchanged since the entry was created, and falls back to put otherwise. This is safe because chain pointers change only together with the map size and every size change bumps the mod count.

Follow-up to #253.

### DIFF
--- a/core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilder.kt
+++ b/core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilder.kt
@@ -86,7 +86,7 @@ internal class PersistentOrderedMapBuilder<K, V>(map: PersistentOrderedMap<K, V>
         }
     }
 
-    internal fun updateLinks(key: K, links: LinkedValue<V>) {
+    internal fun setLinkedValue(key: K, links: LinkedValue<V>) {
         builtMap = null
         hashMapBuilder[key] = links
     }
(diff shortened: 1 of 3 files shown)


core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilder.kt
```diff
@@ -86,6 +86,11 @@ internal class PersistentOrderedMapBuilder<K, V>(map: PersistentOrderedMap<K, V>
         }
     }
 
+    internal fun updateLinks(key: K, links: LinkedValue<V>) {
```
ilya-g: I'd rather name it `setLinkedValue` or `putLinkedValue`

DmitryNekrasov: Done, renamed to `setLinkedValue`.

core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
```diff
@@ -83,13 +83,19 @@ internal class PersistentOrderedMapBuilderEntriesIterator<K, V>(map: PersistentO
...
         get() = links.value
 
     override fun setValue(newValue: V): V {
         val result = links.value
         links = links.withValue(newValue)
-        builder[key] = newValue
+        if (builder.hashMapBuilder.modCount == expectedModCount) {
+            builder.setLinkedValue(key, links)
+        } else {
+            builder[key] = newValue
```
ilya-g, line 97: I'd rather throw `ConcurrentModificationException` here to avoid reinserting already deleted keys for example, but since this part of `MutableMapEntry` behavior is unspecified, we can change it later.

ilya-g (approved):

PR #282, opened 2026-07-16
Add Kover coverage reporting to the core module

DmitryNekrasov:


gradle.properties
```diff
@@ -5,5 +5,6 @@ versionSuffix=SNAPSHOT
 kotlin_version=2.3.0
 dokkaVersion=2.2.0
 kotlinxBenchmarkVersion=0.4.15
+koverVersion=0.9.8
```
fzhinkin: Now you can update it to 0.9.9 :)

core/build.gradle.kts
```diff
@@ -188,6 +189,9 @@ dokka {
 tasks {
     named("jvmTest", Test::class) {
         maxHeapSize = "1024m"
+        providers.gradleProperty("jvmTestExcludes").orNull
```
fzhinkin, line 192: Where this property comes from?

fzhinkin:
Do you plan to setup a baseline rules for coverage [verification](https://kotlin.github.io/kotlinx-kover/gradle-plugin/#verification)?

core/build.gradle.kts
DmitryNekrasov: I added it mainly to be able to run the suite without stress tests:

```
./gradlew jvmTest '-PjvmTestExcludes=tests.stress.*'
```

First, they change the coverage picture, and coverage is not what stress tests are for. Second, they take much longer than the rest of the suite (about 50 seconds against 3). Gradle's `--tests` filter is include-only, so there seems to be no built-in way to exclude tests from the command line.

gradle.properties
DmitryNekrasov: Done

DmitryNekrasov:
> Do you plan to setup a baseline rules for coverage [verification](https://kotlin.github.io/kotlinx-kover/gradle-plugin/#verification)?

Yes, I do, but after I finish working on the coverage and figure out what exact rule I want to apply.

fzhinkin (approved):

PR #285, opened 2026-07-21
Cover persistent list implementation gaps with tests

DmitryNekrasov:
Deterministic tests for the persistent list paths that were exercised only by randomized stress runs or not at all, with sizes at the trie shape boundaries: tail fill, leaf push into the root, height growth.

Coverage of `kotlinx.collections.immutable.implementations.immutableList` by the deterministic suites (stress tests excluded):

|        | Class, %      | Method, %       | Branch, %       | Line, %         | Instruction, %    |
| ------ | ------------- | --------------- | --------------- | --------------- | ----------------- |
| Before | 92.3% (12/13) | 76.3% (103/135) | 49.7% (187/376) | 56.1% (469/836) | 51.1% (2311/4520) |
| After  | 100% (13/13)  | 100% (135/135)  | 92.8% (349/376) | 100% (836/836)  | 98.6% (4458/4520) |

### DIFF
--- a/core/commonTest/src/contract/list/PersistentListBuilderInsertionTest.kt
+++ b/core/commonTest/src/contract/list/PersistentListBuilderInsertionTest.kt
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2016-2026 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package tests.contract.list
+
+import kotlinx.collections.immutable.PersistentList
+import kotlinx.collections.immutable.implementations.immutableList.MAX_BUFFER_SIZE
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.toPersistentList
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class PersistentListBuilderInsertionTest {
+
+    init {
+        check(MAX_BUFFER_SIZE == 32) { "Test sizes assume a trie buffer size of 32" }
+    }
+
+    private fun ownedBuilderOf(size: Int): PersistentList.Builder<Int> {
+        val builder = persistentListOf<Int>().builder()
+        for (element in 0..<size) {
+            assertTrue(builder.add(element))
+        }
+        return builder
+    }
+
+    private fun assertBuilderContents(expected: List<Int>, builder: PersistentList.Builder<Int>) {
+        assertEquals(expected.size, builder.size)
+        assertEquals(expected, builder)
+        for (index in expected.indices) {
+            assertEquals(expected[index], builder[index], "element at index $index")
+        }
+        assertEquals(expected, builder.build())
+    }
+
+    @Test
+    fun `add grows the builder through root creation and root pushes and trie height increases`() {
+        val expected = List(1100) { it }
+        val checkpoints = setOf(33, 64, 65, 97, 1056, 1057, 1089)
+
+        val builder = persistentListOf<Int>().builder()
+        for (element in 0..<1100) {
+            assertTrue(builder.add(element))
+            assertEquals(element + 1, builder.size)
+            if (builder.size in checkpoints) {
+                assertEquals(expected.subList(0, builder.size), builder)
+            }
+        }
+
+        assertBuilderContents(expected, builder)
+    }
+
+    @Test
+    fun `add at an index in a full tail pushes the filled tail into the root as a leaf`() {
+        val builder = ownedBuilderOf(64)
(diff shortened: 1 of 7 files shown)


core/commonTest/src/contract/list/PersistentListBuilderInsertionTest.kt
```diff
@@ -0,0 +1,143 @@
...
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.toPersistentList
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class PersistentListBuilderInsertionTest {
+
+    init {
+        check(MAX_BUFFER_SIZE == 32) { "Test sizes assume a trie buffer size of 32" }
```
fzhinkin: If we decide to change it, for some reason, how to update the test? I mean, it would be nice to mention it in a comment nearby

DmitryNekrasov: If the buffer size ever changes, the tests will fail on this check, and I'd recommend revisiting all the values manually and carefully verifying that every branch of the code under test is still covered. Automatically deriving all the constants in the tests looks less reliable to me, complicates the test logic and makes it less obvious, and isn't worth it: realistically, we are unlikely to ever change `MAX_BUFFER_SIZE`.

fzhinkin: Sure, but right now it's unclear how constants were picked and how to choose new values if `MAX_BUFFER_SIZE` change.

DmitryNekrasov: Done. Documented the picking rules in `checkTrieShapeAssumptions()`'s KDoc.

core/commonTest/src/contract/list/PersistentListBuilderRemovalTest.kt
```diff
@@ -0,0 +1,210 @@
...
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class PersistentListBuilderRemovalTest {
+
+    init {
+        checkTrieShapeAssumptions()
+    }
+
+    private fun ownedBuilderOf(range: IntRange): PersistentList.Builder<Int> {
```
fzhinkin: The lower bound is always `0`, and with that it's essentially the same function as `PersistentListBuilderInsertionTest.ownedBuilderOf`.

It might be worth either extracting it into a base class (or into a separate internal utility function).

core/commonTest/src/contract/list/PersistentVectorTest.kt
```diff
@@ -0,0 +1,161 @@
...
+
+    private fun vectorOfSize(size: Int): PersistentList<Int> {
+        var vector: PersistentList<Int> = persistentListOf()
+        for (element in 0..<size) {
+            vector = vector.adding(element)
+        }
+        return vector
+    }
+
+    private fun assertElementsEqual(expected: List<Int>, actual: List<Int>) {
```
fzhinkin: `kotlin.test.assertContentEquals`?

core/commonTest/src/contract/list/PersistentListBuilderInsertionTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    private fun ownedBuilderOf(size: Int): PersistentList.Builder<Int> {
+        val builder = persistentListOf<Int>().builder()
+        for (element in 0..<size) {
+            assertTrue(builder.add(element))
+        }
+        return builder
+    }
+
+    private fun assertBuilderContents(expected: List<Int>, builder: PersistentList.Builder<Int>) {
```
fzhinkin, line 22: It seems like this one is also can be replaced with `kotlin.test.assertContentEquals`

core/commonTest/src/contract/list/PersistentListBuilderRemovalTest.kt
DmitryNekrasov: Done: extracted `ownedBuilderOf(size)` as an internal top-level function in `listTestUtils.kt`.

core/commonTest/src/contract/list/PersistentVectorTest.kt
DmitryNekrasov: Done, dropped the helper, call sites use `assertContentEquals` directly.

core/commonTest/src/contract/list/PersistentListBuilderInsertionTest.kt
DmitryNekrasov: Done.

PR #287, opened 2026-07-23
Reformat core module sources to the official Kotlin style

DmitryNekrasov:
No code changes: whitespace, line wrapping, and redundant braces only. Small idiom cleanups are split into #288.


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Reformat core module sources to the official Kotlin style
```
fzhinkin: NIce, didn't know about this feature.

fzhinkin (dismissed):

PR #290, opened 2026-07-27
Restore commented-out invariant asserts

DmitryNekrasov:
This PR makes the internal assert zero-cost when disabled, following the kotlinx.coroutines approach: the condition is a lambda that is not evaluated unless assertions are enabled, and each platform provides an inline actual.

- JVM: checks run under `-ea`;
- Native: delegates to `kotlin.assert`;
- JS and Wasm: empty inline actual.

Benchmark runs with assertions disabled show identical allocation profiles and no time regression attributable to the change.

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -108,7 +108,7 @@ internal class TrieNode<K, V>(
     }
 
     /** Returns true if the node bit map has the bit specified by [positionMask] set, indicating there's a subtrie node in the buffer at that position. */
-    private fun hasNodeAt(positionMask: Int): Boolean {
+    internal fun hasNodeAt(positionMask: Int): Boolean {
         return nodeMap and positionMask != 0
     }
 
@@ -123,7 +123,7 @@ internal class TrieNode<K, V>(
     }
 
     /** Retrieves the buffer element at the given [keyIndex] as key of a data entry. */
-    private fun keyAtIndex(keyIndex: Int): K {
+    internal fun keyAtIndex(keyIndex: Int): K {
         @Suppress("UNCHECKED_CAST")
         return buffer[keyIndex] as K
     }
@@ -141,7 +141,7 @@ internal class TrieNode<K, V>(
     }
 
     private fun insertEntryAt(positionMask: Int, key: K, value: V): TrieNode<K, V> {
-//        assert(!hasEntryAt(positionMask))
+        assert { !hasEntryAt(positionMask) }
 
         val keyIndex = entryKeyIndex(positionMask)
         val newBuffer = buffer.insertEntryAtIndex(keyIndex, key, value)
@@ -149,7 +149,7 @@ internal class TrieNode<K, V>(
     }
 
     private fun mutableInsertEntryAt(positionMask: Int, key: K, value: V, owner: MutabilityOwnership): TrieNode<K, V> {
-//        assert(!hasEntryAt(positionMask))
+        assert { !hasEntryAt(positionMask) }
 
         val keyIndex = entryKeyIndex(positionMask)
         if (ownedBy === owner) {
@@ -162,7 +162,7 @@ internal class TrieNode<K, V>(
     }
 
     private fun updateValueAtIndex(keyIndex: Int, value: V): TrieNode<K, V> {
-//        assert(buffer[keyIndex + 1] !== value)
+        assert { buffer[keyIndex + 1] !== value }
 
         val newBuffer = buffer.copyOf()
         newBuffer[keyIndex + 1] = value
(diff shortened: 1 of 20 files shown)


core/jvmMain/src/internal/commonFunctions.kt
```diff
@@ -5,4 +5,14 @@
...
 package kotlinx.collections.immutable.internal
 
-internal actual fun assert(condition: Boolean) = kotlin.assert(condition)
\ No newline at end of file
+private class Assertions
+
+internal val ASSERTIONS_ENABLED = Assertions::class.java.desiredAssertionStatus()
+
+internal actual inline fun assert(condition: () -> Boolean) {
+    if (ASSERTIONS_ENABLED && !condition()) assertionFailed()
```
fzhinkin, line 13: Is there a reason to validate it manually instead of delegating to `kotlin.assert`?

core/nativeMain/src/internal/commonFunctions.kt
```diff
@@ -6,4 +6,6 @@
 package kotlinx.collections.immutable.internal
 
 @OptIn(kotlin.experimental.ExperimentalNativeApi::class)
-internal actual fun assert(condition: Boolean) = kotlin.assert(condition)
+internal actual inline fun assert(condition: () -> Boolean) {
+    kotlin.assert(condition())
```
fzhinkin, line 10: Shall we [enable assertions](https://kotlinlang.org/docs/compiler-reference.html#enable-assertions-ea) for tests?

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -960,7 +960,6 @@ internal class TrieNode<K, V>(
         var nodePositions = nodeMap
         while (nodePositions != 0) {
             val mask = nodePositions.takeLowestOneBit()
-//            assert(hasNodeAt(mask))
```
DmitryNekrasov, line 963: `mask` is `nodePositions.takeLowestOneBit()`, and `nodePositions` starts as `nodeMap` and only has its own bits cleared, so `hasNodeAt(mask)` is true by construction.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -121,7 +121,6 @@ internal class TrieNode<E>(
         newNode: TrieNode<E>,
         owner: MutabilityOwnership?
     ): TrieNode<E> {
-//        assert(buffer[nodeIndex] !== newNode)
```
DmitryNekrasov, line 124: The invariant does not hold on the builder path. When an owned child removes a cell in place, `mutableRemove` falls through with `targetNode === newNode`, so `buffer[nodeIndex] === newNode` is a legal state here and the assert would fire on a routine `builder.remove`.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -205,8 +204,7 @@ internal class TrieNode<E>(
 
 
     private fun removeCellAtIndex(cellIndex: Int, positionMask: Int, owner: MutabilityOwnership?): TrieNode<E> {
-//        assert(!hasNoCellAt(positionMask))
-//        assert(buffer.size > 1) can be false only for the root node
```
DmitryNekrasov, line 209: The root reaches this with `buffer.size == 1` when the last element of a one-element set is removed, so the check cannot be restored as is. A root-gated variant would currently fire too: `retainAll` can leave a single-element node below the root (#291).

core/jvmMain/src/internal/commonFunctions.kt
DmitryNekrasov: Instead of `if (ASSERTIONS_ENABLED && !condition()) assertionFailed()` we can't write `kotlin.assert(condition())`, since in that case `condition()` is evaluated regardless of the `-ea` flag.

core/nativeMain/src/internal/commonFunctions.kt
DmitryNekrasov: Native test binaries are debug, and KGP enables assertions for debuggable binaries.

core/jvmMain/src/internal/commonFunctions.kt
JakeWharton: That can be changed with `-Xassertions=jvm` compiler flag, although it is problematic that it has to be maintained separately.

DmitryNekrasov: I'd prefer to keep the current `if (ASSERTIONS_ENABLED && !condition()) assertionFailed()`: it is more explicit than `-Xassertions=jvm`, and it is consistent with kotlinx.coroutines.

fzhinkin: >  since in that case `condition()` is evaluated regardless of the `-ea` flag.

Should we add (or at least consider doing it) a similar assert function with a lazy condition to the stdlib?

fzhinkin (approved):

core/jvmMain/src/internal/commonFunctions.kt
DmitryNekrasov: Filed KT-88053: https://youtrack.jetbrains.com/issue/KT-88053

JakeWharton: There's already an issue for it: https://youtrack.jetbrains.com/issue/KT-22292

DmitryNekrasov: Indeed, thanks. Closed KT-88053 as a duplicate of KT-22292.

ilya-g (approved):

PR #293, opened 2026-07-28
Lift the single remaining element to the root in set retainAll

### LINKED ISSUE #291: PersistentHashSet.equals is asymmetric after intersect/retainAll
`intersect` (and `PersistentHashSet.Builder.retainAll`) can leave the trie in a non-canonical shape: a single-element node below the root. The `realBitMap == newBitMap` early return in `TrieNode.mutableRetainAll` skips the lift-up of single elements, despite the "single values are kept only on root level" note nearby. Structural equality then depends on the comparison direction, which violates the `equals` contract:

```kotlin
val a = persistentHashSetOf(1, 33) intersect persistentHashSetOf(1, 65)
val b = persistentHashSetOf(1)

a == b // true
b == a // false
```

Removing the remaining element keeps the malformed shape, so the same asymmetry reproduces against the empty set. Size, `contains`, iteration and further modifications behave correctly.

Reproduced on master. Once the shape is fixed, the canonical-form invariant can be asserted in `removeCellAtIndex` (see the discussion in #290).

DmitryNekrasov:
`mutableRetainAll` now does the single-element lift-up before the identity-reuse early return that used to skip it, matching the branch order of `mutableRemoveAll`.

Fixes #291

### DIFF
--- a/core/commonMain/src/implementations/immutableSet/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableSet/TrieNode.kt
@@ -784,6 +784,7 @@ internal class TrieNode<E>(
         }
         // element is directly in buffer
         if (element == buffer[cellIndex]) {
+            assert { shift == 0 || buffer.size > 1 }
             return removeCellAtIndex(cellIndex, cellPositionMask, owner = null)
         }
         return this
@@ -814,6 +815,7 @@ internal class TrieNode<E>(
         }
         // element is directly in buffer
         if (element == buffer[cellIndex]) {
+            assert { shift == 0 || buffer.size > 1 }
             mutator.size--
             return removeCellAtIndex(cellIndex, cellPositionMask, mutator.ownership) // check is empty
         }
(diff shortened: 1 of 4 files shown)


core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -203,8 +203,14 @@ internal class TrieNode<E>(
...
 
-    private fun removeCellAtIndex(cellIndex: Int, positionMask: Int, owner: MutabilityOwnership?): TrieNode<E> {
+    private fun removeCellAtIndex(
+        cellIndex: Int,
+        positionMask: Int,
+        shift: Int,
+        owner: MutabilityOwnership?
+    ): TrieNode<E> {
         assert { !hasNoCellAt(positionMask) }
+        assert { shift == 0 || buffer.size > 1 }
```
ilya-g: Looks strange that `shift` is passed here only to check the predicate of the assertion and is not used otherwise in the function.

DmitryNekrasov: Agree, moved the assert to the two call sites, where shift is already in scope.

ilya-g (approved):

PR #295, opened 2026-07-29
Update Kotlin to 2.4.10

DmitryNekrasov:
The ABI validation DSL changed shape in 2.4.10: invoking abiValidation() is what enables it now, and check depends on the ABI check out of the box, which removes the explicit wiring for KT-78525. The additions in the JVM ABI dump are bridge methods newly emitted by the 2.4 compiler, verified against the published 0.5.0 bytecode; the klib dumps are unchanged.


DmitryNekrasov:
Closing: after folding the Power-Assert wiring into #297, nothing in the assert work needs 2.4 anymore. The upgrade can come back on its own schedule.

Issue #270, opened 2026-06-25
Enable the Power Assert plugin in tests after updating to Kotlin 2.4.20


DmitryNekrasov:
Power Assert shows intermediate values on failed assertions, which would make stress and contract test failures easier to read.

It needs lazy assertion messages, otherwise Power Assert builds a message eagerly for every assertEquals, costly in the stress tests' million-iteration loops. Those overloads landed in [KT-31305](https://youtrack.jetbrains.com/issue/KT-31305) and ship in Kotlin 2.4.20. Once the library is on 2.4.20, enable Power Assert without the kotlin.check workaround that #247 used.

PR #296, opened 2026-07-29
Add opt-in Power-Assert support for invariant asserts

DmitryNekrasov:
With -PpowerAssert, a failed otherwise assert also renders the condition with its sub-expression values:

```
Incorrect size: -1
(size >= 0) otherwise { "Incorrect size: $size" }
 |    |
 -1   false
```

Opt-in only, because this instruments production compilations: transformed bytecode grows at every inlined call site and depends on the experimental kotlin-power-assert-runtime, so it must never reach a published artifact. The override takes compilation default source sets, e.g. -PpowerAssert=jvmMain,jvmTest; the plugin matches exactly those names, hence the computed default instead of commonMain (it would match only the metadata compilation). Unlike #270, which waits for Kotlin 2.4.20 to cover the test suite, this targets the invariant asserts in production sources.

AssertScope.otherwise is added in #297.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,6 +10,7 @@ buildscript {
 
 plugins {
     id("kotlinx.team.infra") version "0.4.0-dev-91"
+    id("org.jetbrains.kotlin.plugin.power-assert") apply false
 }
 
 infra {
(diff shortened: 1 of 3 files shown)


DmitryNekrasov:
Folded into #297: the wiring works the same on 2.3.0, so it no longer needs a separate PR.

PR #299, opened 2026-07-29
Support lazy failure messages in invariant asserts

DmitryNekrasov:
Lets an invariant assert carry a message without paying for it when assertions are off: the condition lambda now takes AssertScope as its receiver, and its infix otherwise attaches a message evaluated only on failure.

```kotlin
assert { (nodeMap == 0) otherwise { "Not a collision node" } }
```

The messages are constants: with -PpowerAssert the diagram prints the values, and the dead branch at each inlined call site stays a string load. Disabled builds still evaluate nothing on any platform.

The Power-Assert plugin is applied only with -PpowerAssert: it instruments production compilations, so the transformed bytecode must never reach a published artifact. The property optionally takes compilation default source sets, e.g. -PpowerAssert=jvmMain,jvmTest.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,6 +10,7 @@ buildscript {
 
 plugins {
     id("kotlinx.team.infra") version "0.4.0-dev-91"
+    id("org.jetbrains.kotlin.plugin.power-assert") apply false
 }
 
 infra {
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -155,6 +155,19 @@ kotlin {
     }
 }
 
+providers.gradleProperty("powerAssert").orNull?.let { requestedSourceSets ->
+    apply(plugin = "org.jetbrains.kotlin.plugin.power-assert")
+    configure<org.jetbrains.kotlin.powerassert.gradle.PowerAssertGradleExtension> {
+        functions.set(listOf("kotlinx.collections.immutable.internal.AssertScope.otherwise"))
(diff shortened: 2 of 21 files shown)


DmitryNekrasov:
How the failures read. With assertions enabled, a broken invariant now fails with its name:

```
java.lang.AssertionError: Inconsistent tail
```

The same failure under -PpowerAssert, with the values filled in by the plugin:

```
java.lang.AssertionError: Inconsistent tail
(size - rootSize(size) <= tail.size.coerceAtMost(MAX_BUFFER_SIZE)) otherwise { "Inconsistent tail" }
 |    | |        |     |  |    |    |            |
 |    | |        |     |  |    |    |            32
 |    | |        |     |  |    |    1
 |    | |        |     |  |    1
 |    | |        |     |  [Ljava.lang.Object;@1728a3fa
 |    | |        |     false
 |    | |        40
 |    | 32
 |    8
 40
```

Another example:

```
java.lang.AssertionError: Buffer overflow
(buffer.size <= MAX_BUFFER_SIZE) otherwise { "Buffer overflow" }
 |      |    |  |
 |      |    |  32
 |      |    false
 |      33
 [Ljava.lang.Object;@66041d36
```

An assert without `otherwise` won't print the sub-expression values under Power-Assert. Тhat would need to be supported separately on the Power-Assert side.

DmitryNekrasov:
@bnorm JFYI.

PR #302, opened 2026-08-10
Take the argument's value when two collision nodes share a key

### LINKED ISSUE #300: putAll keeps the receiver's value for keys shared by two collision nodes
`puttingAll` on persistent hash maps, and `putAll` on their builders, keep the receiver's value for a key that both maps have, when that key sits in a collision node on both sides. `Map.putAll` requires the argument to win.

```kotlin
val a1 = IntWrapper(1, 0)
val a2 = IntWrapper(2, 0)  // full collision with a1

persistentHashMapOf(a1 to 1, a2 to 2).puttingAll(persistentHashMapOf(a1 to 10, a2 to 20))
// {a1=1, a2=2} instead of {a1=10, a2=20}
```

`TrieNode.mutableCollisionPutAll` copies an entry from the argument only when the receiver does not already have that key. It needs at least two keys with the same full hash on each side, so that both maps hold a collision node at the bottom level. Sizes stay correct, only the value is wrong. Ordered maps and a non-persistent argument go through the per-key path and are unaffected.

Found while reviewing #294, but this is the node-vs-node recursion, a different code path.

DmitryNekrasov:
The shortcut that returns the receiver tested that nothing was appended, which stopped meaning nothing changed once a shared key takes the argument's value. Values are compared by `!==`, so a modification means here what it means in `put`.

The merge keeps the stored key instances, as the per-key `put` does, so the argument's collision node is shared only when every shared key is the same instance.

Fixes #300

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -485,18 +485,27 @@ internal class TrieNode<K, V>(
         assert { otherNode.dataMap == 0 }
         val tempBuffer = this.buffer.copyOf(newSize = this.buffer.size + otherNode.buffer.size)
         var i = this.buffer.size
-        for (j in 0..<otherNode.buffer.size step ENTRY_SIZE) {
-            @Suppress("UNCHECKED_CAST")
-            if (!this.collisionContainsKey(otherNode.buffer[j] as K)) {
+        var replaced = false
+        var sharedKeys = true
+        for (j in otherNode.buffer.indices step ENTRY_SIZE) {
+            val keyIndex = this.collisionKeyIndex(otherNode.buffer[j])
+            if (keyIndex == -1) {
                 tempBuffer[i] = otherNode.buffer[j]
                 tempBuffer[i + 1] = otherNode.buffer[j + 1]
                 i += ENTRY_SIZE
-            } else intersectionCounter.count++
+            } else {
+                intersectionCounter.count++
+                if (tempBuffer[keyIndex] !== otherNode.buffer[j]) sharedKeys = false
+                if (tempBuffer[keyIndex + 1] !== otherNode.buffer[j + 1]) {
+                    tempBuffer[keyIndex + 1] = otherNode.buffer[j + 1]
+                    replaced = true
+                }
+            }
         }
 
         return when (val newSize = i) {
-            this.buffer.size -> this
-            otherNode.buffer.size -> otherNode
+            this.buffer.size if !replaced -> this
+            otherNode.buffer.size if sharedKeys -> otherNode
             tempBuffer.size -> TrieNode(0, 0, tempBuffer, owner)
             else -> TrieNode(0, 0, tempBuffer.copyOf(newSize), owner)
         }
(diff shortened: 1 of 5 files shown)


core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -485,18 +485,27 @@ internal class TrieNode<K, V>(
...
+                    replaced = true
+                }
+            }
         }
 
         return when (val newSize = i) {
-            this.buffer.size -> this
-            otherNode.buffer.size -> otherNode
+            this.buffer.size if !replaced -> this
+            otherNode.buffer.size if sharedKeys -> otherNode
```
ilya-g, line 508: Is there a test that checks this behavior? I.e. that original keys are not replaced with the equal but not the same ones after `putAll`

ilya-g: Also, it is ok to use other node in case when it's mutable owned by some other builder and can be changed later?

DmitryNekrasov: Yes, [`putAll that replaces collision values should keep the stored key instances`](https://github.com/Kotlin/kotlinx.collections.immutable/blob/5bc568ca35aff3d1da74e806943267e370e430ca/core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt#L154-L163).

DmitryNekrasov: It is ok. For a builder argument the `putAll` method calls the `build` method first, and the `build` method creates a new ownership, so no node of the argument stays owned by a builder that can change it later.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -671,6 +671,7 @@ internal class TrieNode<K, V>(
         intersectionCounter: DeltaCounter,
         mutator: PersistentHashMapBuilder<K, V>
     ): TrieNode<K, V> {
+        assert { otherNode.ownedBy !== mutator.ownership }
```
ilya-g: This assert would pass if `otherNode` is still owned by other mutator. Perhaps it should check if otherNode is not owned or owned by the same mutator.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
DmitryNekrasov: Also, I have added an assert for this.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
ilya-g: Or it is not possible to check if a non-null ownership allows mutation or is already lost after calling `build`?

DmitryNekrasov: Yes, you are right, thank you!

I would propose a flag in `MutabilityOwnership` that the `build` method sets, and I would do that separately, since the list and the set builders have the same behaviour. WDYT?

DmitryNekrasov: I don't think the check method you suggested will work.

ilya-g: Yes, I also think that without some flag indicating that ownership is active it would not be possible to assert.

ilya-g (approved):

PR #312, opened 2026-08-14
putAll that only replaces values leaves a live builder iterator on the old values

### LINKED ISSUE #308: putAll that only replaces values leaves a live builder iterator on the old values
`putAll` on a persistent hash map builder replaces the values, but an iterator taken before the call keeps returning the ones from before it, and never throws. The same call throws `ConcurrentModificationException` as soon as it also adds a key.

To reproduce:

```kotlin

builder.putAll(persistentHashMapOf(1 to "x", 2 to "y"))

builder[1]             // x
iterator.next().value  // a, expected either x or ConcurrentModificationException
```

Keys that share a hash code behave the same way, so this is not specific to collision nodes.

DmitryNekrasov:
The fast-path merge replaces values without moving `modCount`, so the builder counted adds and removals but not the overwrites. While fixing, I noticed the merge also writes the argument's key over an equal stored key, which is a node replacement no value check can see, so the entry cell now keeps the stored key, as the per-key `put` does.

Fixes #308

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -480,7 +480,7 @@ internal class TrieNode<K, V>(
     private fun mutableCollisionPutAll(
         otherNode: TrieNode<K, V>,
         intersectionCounter: DeltaCounter,
-        owner: MutabilityOwnership
+        mutator: PersistentHashMapBuilder<K, V>
     ): TrieNode<K, V> {
         assert { nodeMap == 0 }
         assert { dataMap == 0 }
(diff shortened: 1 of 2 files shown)


core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -723,12 +725,22 @@ internal class TrieNode<K, V>(
...
+                    val otherKeyIndex = otherNode.entryKeyIndex(positionMask)
+                    val otherValue = otherNode.valueAtKeyIndex(otherKeyIndex)
+                    if (this.hasEntryAt(positionMask)) {
+                        val thisKeyIndex = this.entryKeyIndex(positionMask)
+                        intersectionCounter.count++
+                        if (mutableNode !== this) {
+                            val thisValue = this.valueAtKeyIndex(thisKeyIndex)
+                            if (thisValue !== otherValue) mutator.modCount++
+                        }
+                        mutableNode.buffer[newKeyIndex] = this.keyAtIndex(thisKeyIndex)
```
ilya-g, line 739: Is this a no-op when `mutableNode === this` or `newKeyIndex` can be different from `thisKeyIndex`, so the key needs to be moved in buffer?

ilya-g (dismissed):

ilya-g (approved):
see previous comment

core/commonMain/src/implementations/immutableMap/TrieNode.kt
DmitryNekrasov: Yes, in this case it's a no-op. `mutableNode === this` requires `newDataMap == this.dataMap`, and with equal bitmaps `newKeyIndex == thisKeyIndex`. They can differ only for a newly allocated `mutableNode`, where this line fills the new buffer.

ilya-g: Then this line can be moved inside `if (mutableNode !== this) {` condition together with `thisKeyIndex` initialization

PR #315, opened 2026-08-17
setValue on an ordered map builder entry re-adds a key that was already removed

### LINKED ISSUE #310: setValue on an ordered map builder entry re-adds a key that was already removed
An entry taken from `persistentMapOf(...).builder().entries` keeps writing to the map after its key is gone. `setValue` on it puts the key back, at the end of the iteration order, and the size grows with it. The builder of `persistentHashMapOf` returns without touching the map, and so does `LinkedHashMap`.

To reproduce:

```kotlin
val entry = iterator.next()
iterator.remove()

entry.setValue("z")

builder.size                   // 2, expected 1
builder.build().keys.toList()  // [2, 1], expected [2]
```

`builder.remove(1)` in place of `iterator.remove()` gives the same result.

Found while fixing #307.

DmitryNekrasov:
The write of a stale entry's `setValue` went through `put`, so for a removed key it turned into an insert at the tail of the chain. The fix is to write through only while the key is still stored, the same check the hash builder makes.

Fixes #310

### DIFF
--- a/core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
+++ b/core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
@@ -96,7 +96,7 @@ private class MutableMapEntry<K, V>(
         links = links.withValue(newValue)
         if (builder.hashMapBuilder.sizeModCount == expectedModCount) {
             builder.setLinkedValue(key, links)
-        } else {
+        } else if (builder.containsKey(key)) {
             builder[key] = newValue
         }
         return result
--- a/core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
+++ b/core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
@@ -374,4 +374,21 @@ class PersistentHashMapBuilderTest {
         assertEquals("x", builder[storedKey])
         assertSame(storedKey, builder.keys.single { it == storedKey })
     }
+
+    @Test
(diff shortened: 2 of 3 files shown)


core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
```diff
@@ -96,7 +96,7 @@ private class MutableMapEntry<K, V>(
         links = links.withValue(newValue)
         if (builder.hashMapBuilder.sizeModCount == expectedModCount) {
             builder.setLinkedValue(key, links)
-        } else {
+        } else if (builder.containsKey(key)) {
             builder[key] = newValue
```
ilya-g, line 100: Perhaps we could use `replace(key, newValue)` (https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#replace-K-V-) to do it in one key lookup if we had this operation in `MutableMap` interface or provided it internally, but I suppose since this is a defensive fallback for an unspecified behavior, it isn't worth much to optimize this path.

ilya-g (approved):

core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
DmitryNekrasov: Agree. The one-lookup version through `setLinkedValue` would also drop the built-map cache even for the stored value.

Issue #305, opened 2026-08-10
Consider property-based tests for the hash trie


DmitryNekrasov:
The random tests in `core/commonTest/src/stress` never build the tries where the recent bugs were. `WrapperGenerator` takes the hashes from a range that depends on the element count, so the high bits are always zero and a collision node at `MAX_SHIFT` never appears. #294, #300, #304, #313, #316 and #322 all passed these tests.

I propose to generate the keys from a description of the trie: which keys share a path, where they split, which ones collide. A failing case becomes a small regression test. We can look at kotlinx.fuzz or jqwik.

PR #317, opened 2026-08-18
puttingAll, addingAll and retainingAll replace the stored key or element instance when one side holds it in a deeper node

### LINKED ISSUE #313: puttingAll replaces the stored key instance when the argument holds the key in a deeper node
When both maps hold the key as a direct entry, `puttingAll` keeps the stored key and takes the argument's value, the way a single `putting` call does. But when the argument holds the same key deeper, together with another key that shares a hash-code segment, the stored key instance is replaced by the argument's equal one. `java.util` maps keep the stored key in every case.

To reproduce:

```kotlin
val storedKey = IntWrapper(1, 0)
val map = persistentHashMapOf(storedKey to "a")

val updated = map.puttingAll(persistentHashMapOf(IntWrapper(1, 0) to "x", IntWrapper(2, 32) to "y"))

updated.keys.single { it == storedKey } === storedKey  // false, expected true
```

The same happens when the argument stores the key in a bottom-level collision node, e.g. with `IntWrapper(2, 0)` as the second key.

### LINKED ISSUE #316: addingAll replaces the stored element instance when the argument holds the element in a deeper node
When both sets hold the element as a direct entry, `addingAll` keeps the stored element instance, the way a single `adding` call does. But when the argument holds the equal element deeper, together with another element that shares a hash-code segment, the stored instance is replaced by the argument's equal one. `java.util` sets keep the stored element in every case.

To reproduce:

```kotlin
val set = persistentHashSetOf(storedElement)

val updated = set.addingAll(persistentHashSetOf(IntWrapper(1, 0), IntWrapper(2, 32)))

```

The same happens when the argument stores the element in a bottom-level collision node, e.g. with `IntWrapper(2, 0)` as the second element. And when both sides store their elements in collision nodes and the receiver's elements are a subset of the argument's, every stored instance is replaced at once:

```kotlin
val set = persistentHashSetOf(storedElement, IntWrapper(2, 0))

val updated = set.addingAll(persistentHashSetOf(IntWrapper(1, 0), IntWrapper(2, 0), IntWrapper(3, 0)))

```

The map's `puttingAll` had the same class of problem, fixed in #302 and #313.

### LINKED ISSUE #322: retainingAll replaces the stored element instance when the receiver holds the element in a deeper node
When both sets hold the element as a direct entry, `retainingAll` keeps the stored element instance. But when the receiver holds the equal element deeper, together with another element that shares a hash-code segment, the survivor is the argument's equal instance. `java.util` sets keep the stored element in every case.

To reproduce:

```kotlin
val set = persistentHashSetOf(storedElement, IntWrapper(2, 32))

val updated = set.retainingAll(persistentHashSetOf(IntWrapper(1, 0)))

```

And when both sides store their elements in collision nodes and the argument's elements are a subset of the receiver's, every retained instance is replaced at once:

```kotlin
val set = persistentHashSetOf(storedElement, IntWrapper(2, 0), IntWrapper(3, 0))

val updated = set.retainingAll(persistentHashSetOf(IntWrapper(1, 0), IntWrapper(2, 0)))

```

The set's `addingAll` had the same class of problem, fixed in #321.

DmitryNekrasov:
Since #312 every cell of the map's bulk merge keeps the stored key, except the one where the receiver holds the key as an entry and the argument holds it in a subtree, which took the argument's subtree as is. The fix lifts the entry into a one-entry node, so the merge fixed in #312 handles this cell too.

`addingAll` on the set loses the stored element in the same cell, `retainingAll` in the opposite one, where the receiver holds the subtree. Both get the collision guard from #302. The `addingAll` fix repeats the map's, but the lifted node stays unowned, since with the set's single bitmap an owned one is reused in place down a collision chain and the merge returns it instead of the argument's node. For `retainingAll` a lookup is enough, since a subtree and one element have at most that element in common.

Fixes #313
Fixes #316
Fixes #322

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableMap/TrieNode.kt
@@ -561,28 +561,22 @@ internal class TrieNode<K, V>(
             val otherTargetNode = otherNode.nodeAtIndex(otherNode.nodeIndex(positionMask))
             when {
                 this.hasEntryAt(positionMask) -> {
-                    // if otherTargetNode already has a value associated with the key, do not put this entry
+                    // keep this node's key if the subtree also holds it, and take the argument's value
                     val keyIndex = this.entryKeyIndex(positionMask)
                     val key = this.keyAtIndex(keyIndex)
-                    val hasKey = if (shift == MAX_SHIFT) {
-                        otherTargetNode.collisionContainsKey(key)
+                    val value = this.valueAtKeyIndex(keyIndex)
+                    val entryNode = if (shift == MAX_SHIFT) {
+                        TrieNode<K, V>(0, 0, arrayOf(key, value), mutator.ownership)
                     } else {
-                        otherTargetNode.containsKey(key.hashCode(), key, shift + LOG_MAX_BRANCHING_FACTOR)
+                        val keyPositionMask = 1 shl indexSegment(key.hashCode(), shift + LOG_MAX_BRANCHING_FACTOR)
+                        TrieNode(keyPositionMask, 0, arrayOf(key, value), mutator.ownership)
                     }
-                    if (hasKey) {
-                        intersectionCounter.count++
-                        otherTargetNode
-                    } else {
-                        val value = this.valueAtKeyIndex(keyIndex)
-                        if (shift == MAX_SHIFT) {
-                            otherTargetNode.mutableCollisionPut(key, value, mutator)
-                        } else {
-                            otherTargetNode.mutablePut(
-                                key.hashCode(), key, value,
-                                shift + LOG_MAX_BRANCHING_FACTOR, mutator
-                            )
-                        }
-                    }
+                    entryNode.mutablePutAll(
+                        otherTargetNode,
+                        shift + LOG_MAX_BRANCHING_FACTOR,
+                        intersectionCounter,
+                        mutator
+                    )
                 }
 
                 else -> otherTargetNode
--- a/core/commonMain/src/implementations/immutableSet/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableSet/TrieNode.kt
@@ -47,15 +47,14 @@ private fun Array<Any?>.removeCellAtIndex(cellIndex: Int): Array<Any?> {
 }
 
 /**
- * Writes all elements from [this] to [newArray], starting with [newArrayOffset], filtering
- * on the fly using [predicate]. By default, filters out [TrieNode.EMPTY] instances
+ * Writes all elements from [this] to [newArray], filtering on the fly using [predicate].
+ * By default, filters out [TrieNode.EMPTY] instances
  *
  * return number of elements written to [newArray]
  **/
 @IgnorableReturnValue
 private inline fun Array<Any?>.filterTo(
     newArray: Array<Any?>,
-    newArrayOffset: Int = 0,
     predicate: (Any?) -> Boolean = { it !== TrieNode.EMPTY }
 ): Int {
     var i = 0
@@ -64,9 +63,9 @@ private inline fun Array<Any?>.filterTo(
         assert { j <= i } // this is extremely important if newArray === this
         val e = this[i]
         if (predicate(e)) {
-            newArray[newArrayOffset + j] = this[i]
+            newArray[j] = this[i]
             ++j
-            assert { newArrayOffset + j <= newArray.size }
+            assert { j <= newArray.size }
         }
         ++i
     }
@@ -265,17 +264,26 @@ internal class TrieNode<E>(
             return this
         }
         val tempBuffer = this.buffer.copyOf(newSize = this.buffer.size + otherNode.buffer.size)
-        val totalWritten = otherNode.buffer.filterTo(tempBuffer, newArrayOffset = this.buffer.size) {
+        var totalSize = this.buffer.size
+        var sharedElements = true
+        for (j in otherNode.buffer.indices) {
             @Suppress("UNCHECKED_CAST")
-            !this.collisionContainsElement(it as E)
+            val otherElement = otherNode.buffer[j] as E
+            val index = this.buffer.indexOf(otherElement)
+            if (index == -1) {
+                tempBuffer[totalSize] = otherElement
+                totalSize++
+            } else {
+                intersectionSizeRef.count++
+                if (tempBuffer[index] !== otherElement) sharedElements = false
+            }
         }
-        val totalSize = totalWritten + this.buffer.size
-        intersectionSizeRef += (tempBuffer.size - totalSize)
-        if (totalSize == this.buffer.size) return this
-        if (totalSize == otherNode.buffer.size) return otherNode
-
-        val newBuffer = if (totalSize == tempBuffer.size) tempBuffer else tempBuffer.copyOf(newSize = totalSize)
-        return setProperties(newBitmap = 0, newBuffer, owner)
+        return when (totalSize) {
+            this.buffer.size -> this
+            otherNode.buffer.size if sharedElements -> otherNode
+            tempBuffer.size -> setProperties(newBitmap = 0, newBuffer = tempBuffer, owner)
+            else -> setProperties(newBitmap = 0, newBuffer = tempBuffer.copyOf(newSize = totalSize), owner)
+        }
     }
 
     private fun mutableCollisionRetainAll(
@@ -393,7 +401,8 @@ internal class TrieNode<E>(
         // for each bit set in the resulting mask,
         // either left, right or both masks contain the same bit
         // Note: shift can be MAX_SHIFT here, where a cell holding a node holds a collision node,
-        // so the node-vs-element branches dispatch to the collision helpers instead of descending
+        // so node-vs-element dispatches to the collision helper, and element-vs-node lifts the element
+        // into a bitmap-0 node whose mutableAddAll call past MAX_SHIFT routes to mutableCollisionAddAll
         newBitMap.forEachOneBit { positionMask, newNodeIndex ->
             val thisIndex = indexOfCellAt(positionMask)
             val otherNodeIndex = otherNode.indexOfCellAt(positionMask)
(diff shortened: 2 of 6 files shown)


qwwdfsad:
Could you please elaborate on why it's important to have (find, fix and test for) this contract? Is it documented anywhere?

DmitryNekrasov:
@qwwdfsad 
> Could you please elaborate on why it's important to have (find, fix and test for) this contract? Is it documented anywhere?

The `puttingAll` KDoc states the call is equivalent to calling `put(k, v)` once for each mapping, and `put` keeps the stored key instance. However, you can observe it only through `===` on keys, so I set the priority to low. If we want this as an explicit contract, I can mention it in KDoc, WDYT?

qwwdfsad:
>and put keeps the stored key instance

We don't have it in [put](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-mutable-map/put.html) contract either. Thus, I was surprised this bug was found and fixed, but no mentions, bugs or investigations were filed against `MutableMap`, though it's the first obvious place to look when thinking about this problem

DmitryNekrasov:
> We don't have it in put contract either.

I agree, this behavior is not documented explicitly, but it is the actual behavior (I checked `HashMap` and `LinkedHashMap` on all platforms), and in the `puttingAll` KDoc we promise behavior equivalent to `put`.

qwwdfsad:
Yes, but the implementation shouldn't drive the solution.

We can:

* Pretend such a contract exists implicitly based on a few implementations in the wild
* Ignore anything about the contract and just say we are fixing this because LHM/HM does that
* Notice that it's not an actual contract and make an explicit decision about it (either it **is** a contract that should be documented or it isn't, then we probably can even get away with not fixing that), filing a ticket, fixing the documentation in stdlib, etc.

I would propose to go with the last one

DmitryNekrasov:
> I would propose to go with the last one

I propose to file a ticket for adding this contract to the stdlib. Also, keeping the stored key agrees with what the other shapes of this merge already do. When both maps hold the key as a direct entry, when both hold it in bottom-level collision nodes, and when the receiver holds it inside a subtree while the argument brings it as an entry, the stored key survives. The shape fixed here was the only one that replaced it.

qwwdfsad:
Sure, go ahead

DmitryNekrasov:
> Sure, go ahead

[KT-88708](https://youtrack.jetbrains.com/issue/KT-88708) Consider documenting that MutableMap.put keeps the stored key instance

fzhinkin:
This PR, #321, and #328 all solving exactly the same problem, but for different functions. What's the reason to split these changes between separate PRs? 

It would be much easier for everyone to group similar changes together (note that I don't suggest squashing commits) and review them all at once (easier in terms of switching and restoring the context of the problem and checking how the same issue was solved in other PR for the reviewer, and addressing the same review comment in all PRs for you).

fzhinkin:
The same (https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533) applies to other PRs in the stack, like those about checking for commodification in `Iterator.remove`.

In general, stacked PRs are nice when there are several dependent (and large-scale) changes that need separate reviews. Spitting similar small and related changes that could and should be reviewed together across multiple PRs is not the best way to use this feature. In this case, it only adds additional overhead.

DmitryNekrasov:
Done. #321 and #328 are here as separate commits. The same for the rest of the stack: #320, #329, #334 and #336 went into #324, #343 into #339, #346 into #337. #341, #347, #349 and #350 stay separate, since each of them fixes a different problem.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
```diff
@@ -561,28 +561,22 @@ internal class TrieNode<K, V>(
...
                 this.hasEntryAt(positionMask) -> {
-                    // if otherTargetNode already has a value associated with the key, do not put this entry
+                    // keep this node's key if the subtree also holds it, and take the argument's value
                     val keyIndex = this.entryKeyIndex(positionMask)
                     val key = this.keyAtIndex(keyIndex)
-                    val hasKey = if (shift == MAX_SHIFT) {
-                        otherTargetNode.collisionContainsKey(key)
+                    val value = this.valueAtKeyIndex(keyIndex)
+                    val entryNode = if (shift == MAX_SHIFT) {
+                        TrieNode<K, V>(0, 0, arrayOf(key, value), mutator.ownership)
```
fzhinkin, line 569: In this case, `mutablePutAll` will end up creating yet another node and re-allocating buffer array. Knowing that, perhaps we can adapt / rewrite `mutableCollisionPutAll` to avoid unnecessary allocations? Or it is not worth it, in general?

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
```diff
@@ -375,6 +376,220 @@ class PersistentHashMapBuilderTest {
...
+
+        assertEquals(3, builder.size)
+        assertEquals("a", builder[storedKey])
+        assertEquals("y", builder[IntWrapper(2, 32)])
+        assertEquals("z", builder[IntWrapper(3, 64)])
+        assertSame(storedKey, builder.keys.single { it == storedKey })
+    }
+
+    @Test
+    fun `putAll should push the entry deeper when it collides with an argument entry inside the subtree`() {
```
fzhinkin, line 474: The same set of concerns I had regarding tests from  https://github.com/Kotlin/kotlinx.collections.immutable/pull/285 applies to many tests added here.

Perhaps, instead of writing such tests manually, it makes sense to model relationship between different nodes using some high-level API, and then generate different arrangements of nodes using it?

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
```diff
@@ -375,6 +376,220 @@ class PersistentHashMapBuilderTest {
...
+        val storedKey = IntWrapper(1, 0)
+        val builder = persistentHashMapOf(storedKey to "a").builder()
+
+        builder.putAll(persistentHashMapOf(IntWrapper(2, 1 shl 10) to "y", IntWrapper(3, 32) to "z"))
```
fzhinkin: It would be nice to briefly elaborate why 32 and 1024. And corresponding wrappers could be reused.

core/commonMain/src/implementations/immutableSet/TrieNode.kt
```diff
@@ -434,21 +462,21 @@ internal class TrieNode<E>(
...
+                                TrieNode<E>(0, arrayOf(thisCell), null)
                             } else {
-                                otherNodeCell.mutableAdd(
-                                    thisCell.hashCode(), thisCell,
-                                    shift + LOG_MAX_BRANCHING_FACTOR, mutator
-                                )
-                            }.also {
-                                if (mutator.size == oldSize) intersectionSizeRef.count++
+                                val elementPositionMask =
+                                    1 shl indexSegment(thisCell.hashCode(), shift + LOG_MAX_BRANCHING_FACTOR)
```
fzhinkin, line 473: Thinking out loud: it might be the time to replace `1 shl indexSegment(hc, shift)` with `positionMask(hc, shift)` or something like that.

DmitryNekrasov: Agree, but let's do it as a separate PR (#360).

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
DmitryNekrasov: Done. The wrappers are now shared keys in `TrieKeys.kt`, derived from `LOG_MAX_BRANCHING_FACTOR` and `MAX_SHIFT`, with a short comment on how a hash maps to a level and a cell.

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
DmitryNekrasov: Agree, and there is a separate task for it (#305). I have moved the constants to `TrieKeys.kt`. But I would keep the hand-written tests here as well, since each of them covers one case and a failure is easy to read.

core/commonMain/src/implementations/immutableMap/TrieNode.kt
DmitryNekrasov: I think it is not worth it. This branch runs only when two keys share all 32 bits of the hash, the rarest case of the merge.

fzhinkin: BTW, what's the estimated performance impact of all PRs aligning behavior with the contracts for corresponding interfaces?

core/commonTest/src/trie/TrieKeys.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package tests.trie
+
+import kotlinx.collections.immutable.implementations.immutableMap.LOG_MAX_BRANCHING_FACTOR
+import kotlinx.collections.immutable.implementations.immutableMap.MAX_SHIFT
+import tests.IntWrapper
+
+// Keys for the hash trie tests. A key with hash `c shl (n * LOG_MAX_BRANCHING_FACTOR)` shares the path of the
+// hash 0 keys for n levels and sits in cell c of the level n node, next to them.
+val a1 = IntWrapper(1, 0)
```
fzhinkin: These names (a1, a2, a3) kinda worked for local variables in tests, but they are definitely not the best choice for top-level properties shared across multiple tests.

core/commonTest/src/trie/TrieKeys.kt
```diff
@@ -0,0 +1,21 @@
...
+val a2 = IntWrapper(2, 0)
+val a3 = IntWrapper(4, 0)
+val sibling = IntWrapper(3, 1 shl MAX_SHIFT)
```
fzhinkin: Whose sibling is that?

core/commonTest/src/trie/TrieKeys.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright 2016-2026 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package tests.trie
+
+import kotlinx.collections.immutable.implementations.immutableMap.LOG_MAX_BRANCHING_FACTOR
+import kotlinx.collections.immutable.implementations.immutableMap.MAX_SHIFT
+import tests.IntWrapper
+
```
fzhinkin, line 11: If implementing tests bound to internal implementation of tested data structures is inevitable, and there are many other tests to write, did you consider providing a DSL to both make tests less obscure and simplify creation of differently shaped tries in future tests?

I guess, a DSL where you can explicitly declare a trie with required properties and relationships between nodes and then translate it into a map/set, would do the trick.

core/commonTest/src/trie/TrieKeys.kt
DmitryNekrasov: Done, renamed to `collidingKeyX`.

core/commonTest/src/trie/TrieKeys.kt
DmitryNekrasov: Agree, it made sense in the previous place, but not now, I renamed it to `lastLevelSibling`.

core/commonTest/src/trie/TrieKeys.kt
DmitryNekrasov: That sounds like a very good idea, thanks! However, I wasn't able to quickly design such an API on a first pass. The task seems substantial. Given that these tests are minimal and pin specific contracts, and that we’ve found many bugs and want to get fixes to users ASAP, I would suggest to design such a DSL and working on it in parallel with the rest of the library stabilization. WDYT?

core/commonMain/src/implementations/immutableMap/TrieNode.kt
DmitryNekrasov: Two observations. First, these changes improve correctness. Second, none of the changes affect the time complexity on hot paths. They do the same work plus a few `modCount` checks, and the new allocations are only on rare merge scenarios. Where an iterator or an entry re-reads its slot after an external modification (#339, #341), it costs one more descent of the trie, and only after the modification. I have not run benchmarks for these changes, but I think it could be done, following the style @ilya-g used recently, where benchmarks are based on tests. However, I don't think we'll see any significant difference.

core/commonTest/src/trie/TrieKeys.kt
fzhinkin: There are tons of tests added in other PRs. Many of them creates a trie with a particular shape. Others, do something else but, in general, could be reduced to a few particular test patterns that are currently just copy-pasted over and over with minimal difference instead of abstracting the common boilerplate code out, leaving only meaningful part to the test.
For me, it seems more productive to create a test infrastructure to write concise and simple tests that are easier to review and maintain (and write, if that's a concern) first. Doing it the other way around mean that at least reviewers will have to spend more time reviewing tests, figuring out what they are supposed to do, and all that only to review these exact tests once again after you'll design the DSL.

Issue #319, opened 2026-08-19
remove on a hash map builder iterator throws an AssertionError after a remove of a different key


DmitryNekrasov:
`remove()` on an entries iterator of `persistentHashMapOf(...).builder()` throws an `AssertionError` from inside the library after some other key is removed directly from the builder. `LinkedHashMap` throws `ConcurrentModificationException` here.

To reproduce:

```kotlin
val builder = persistentHashMapOf(1 to "a", 2 to "b").builder()
val iterator = builder.entries.iterator()
iterator.next()

builder.remove(2)

iterator.remove()  // AssertionError
```

Found while fixing #314.

PR #320, opened 2026-08-19
setValue on a hash map builder entry throws an AssertionError after a remove of a different key

### LINKED ISSUE #314: setValue on a hash map builder entry throws an AssertionError after a remove of a different key
`setValue` on an entry taken from `persistentHashMapOf(...).builder().entries` throws an `AssertionError` from inside the library after some other key is removed directly from the builder. `LinkedHashMap` accepts the write, and the following `next()` throws `ConcurrentModificationException`. The builder of `persistentMapOf` behaves like `LinkedHashMap` here.

To reproduce:

```kotlin
val entry = iterator.next()

builder.remove(2)

entry.setValue("z")  // AssertionError
```

Found while fixing #310.

DmitryNekrasov:
`setValue` never compared `builder.modCount` to `expectedModCount` before using the cached path, so after an external remove `resetPath` descended to a key that is gone, and when the path's key survived, the trailing resync swallowed the comodification. The fix is the ordered entry's dirty branch from #315. While fixing, I noticed `iterator.remove()` has the same stale descent - filed as #319.

Fixes #314

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
@@ -73,6 +73,11 @@ internal open class PersistentHashMapBuilderBaseIterator<K, V, T>(
     fun setValue(key: K, newValue: V) {
         if (!builder.containsKey(key)) return
 
+        if (builder.modCount != expectedModCount) {
+            builder[key] = newValue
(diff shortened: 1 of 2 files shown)


DmitryNekrasov:
I closed and opened the PR several times because it wasn't linked to the task, I hoped it would help, but it hasn't yet.

DmitryNekrasov:

PR #321, opened 2026-08-19
addingAll replaces the stored element instance when the argument holds the element in a deeper node

### LINKED ISSUE #316: addingAll replaces the stored element instance when the argument holds the element in a deeper node

DmitryNekrasov:
The fix mirrors #317's lift-and-merge and the #302 guard. However, unlike the map's, the lifted node stays unowned, since with the set's single bitmap an owned lift gets reused in place down a collision chain, and then the merge returns the lift instead of adopting the argument's untouched node.

Fixes #316

### DIFF
--- a/core/commonMain/src/implementations/immutableSet/TrieNode.kt
+++ b/core/commonMain/src/implementations/immutableSet/TrieNode.kt
@@ -265,17 +265,26 @@ internal class TrieNode<E>(
             return this
         }
         val tempBuffer = this.buffer.copyOf(newSize = this.buffer.size + otherNode.buffer.size)
-        val totalWritten = otherNode.buffer.filterTo(tempBuffer, newArrayOffset = this.buffer.size) {
+        var totalSize = this.buffer.size
+        var sharedElements = true
+        for (j in otherNode.buffer.indices) {
             @Suppress("UNCHECKED_CAST")
-            !this.collisionContainsElement(it as E)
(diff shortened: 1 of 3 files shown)


DmitryNekrasov:
Moved to #317 as a separate commit, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #324, opened 2026-08-20
remove on builder iterators and setValue on a hash map builder entry do not see a remove of a different key through the builder

### LINKED ISSUE #314: setValue on a hash map builder entry throws an AssertionError after a remove of a different key

### LINKED ISSUE #319: remove on a hash map builder iterator throws an AssertionError after a remove of a different key

### LINKED ISSUE #325: remove on a hash set builder iterator throws an AssertionError after a remove of a different element
`remove()` on an iterator of `persistentHashSetOf(...).builder()` throws an `AssertionError` from inside the library after some other element is removed directly from the builder. `LinkedHashSet` throws `ConcurrentModificationException` here.

To reproduce:

```kotlin
iterator.next()

builder.remove(2)

iterator.remove()  // AssertionError
```

The hash map twin is #319.

### LINKED ISSUE #326: remove on an ordered map builder iterator succeeds after a remove of a different key
`remove()` on an entries iterator of `persistentMapOf(...).builder()` succeeds after some other key is removed directly from the builder, where `LinkedHashMap` throws `ConcurrentModificationException`.

To reproduce:

```kotlin
iterator.next()

builder.remove(2)

```

With 3 entries and an already visited key removed, an entry is also lost from the iteration:

```kotlin
iterator.next()
iterator.next()

builder.remove(1)

iterator.hasNext()  // false, so the entry for 3 is never returned
```

`LinkedHashMap` keeps both 2 and 3 there, since it throws instead of removing.

With 3 entries and the upcoming key removed, `iterator.remove()` also succeeds, and the following `next()` throws `ConcurrentModificationException` with a message that the hash code of a key has changed, which is misleading.

The hash map twin is #319.

### LINKED ISSUE #327: remove on an ordered set builder iterator succeeds after a remove of a different element
`remove()` on an iterator of `persistentSetOf(...).builder()` succeeds after some other element is removed directly from the builder, where `LinkedHashSet` throws `ConcurrentModificationException`.

To reproduce:

```kotlin
val builder = persistentSetOf(1, 2).builder()
iterator.next()

builder.remove(2)

```

With 3 elements and an already visited one removed, an element is also lost from the iteration:

```kotlin
iterator.next()
iterator.next()

builder.remove(1)

iterator.hasNext()  // false, so 3 is never returned
```

`LinkedHashSet` keeps both 2 and 3 there, since it throws instead of removing.

The ordered map twin is #326.

DmitryNekrasov:
`setValue` on the hash map builder's iterator used the cached path without comparing `builder.modCount` to `expectedModCount`, so after a remove through the builder it went down to a key that is gone, and when the key was still there, the update of `expectedModCount` at the end hid the comodification. The fix is to write through the builder when `modCount` has moved, as the ordered entry does.

Also, `remove()` on all 4 builder iterators had no comodification check, where `next()` has one.

Fixes #314
Fixes #319
Fixes #325
Fixes #326
Fixes #327


core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
```diff
@@ -777,4 +777,44 @@ class PersistentHashMapBuilderTest {
         assertEquals(order[2], iterator.next().key)
         assertFalse(iterator.hasNext())
     }
+
+    @Test
+    fun `iterator remove after a remove of a different key throws ConcurrentModificationException`() {
+        val builder = persistentHashMapOf(1 to "a", 2 to "b").builder()
+        val order = builder.keys.toList()
```
fzhinkin: Nit: `orderedKeys` or just `keys`?

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
```diff
@@ -777,4 +777,44 @@ class PersistentHashMapBuilderTest {
...
+    @Test
+    fun `iterator remove on the exhausted iterator of a single entry map after an external put throws ConcurrentModificationException`() {
+        val builder = persistentHashMapOf(1 to "a").builder()
+        val iterator = builder.entries.iterator()
+        val _ = iterator.next()
+        assertFalse(iterator.hasNext())
+
+        builder[2] = "b"
+
+        assertFailsWith<ConcurrentModificationException> { iterator.remove() }
```
fzhinkin, line 813: A similar check for `next`'s behavior is also needed here.

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
```diff
@@ -777,4 +777,44 @@ class PersistentHashMapBuilderTest {
...
+    }
+
+    @Test
+    fun `iterator remove without a preceding next after an external remove throws IllegalStateException`() {
+        val builder = persistentHashMapOf(1 to "a", 2 to "b").builder()
+        val iterator = builder.entries.iterator()
+
+        builder.remove(1)
+
+        assertFailsWith<IllegalStateException> { iterator.remove() }
```
fzhinkin, line 826: ISE has no message here, whilst it should say something like, "next call is required".

fzhinkin (changes requested):

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
DmitryNekrasov: Renamed to `orderedKeys` here and in the rest of the stack, `orderedElements` in the set tests.

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
DmitryNekrasov: Added. The same row went into the hash set and ordered set twins (#329, #336), which have the same exhausted-iterator test.

core/commonTest/src/contract/map/PersistentHashMapBuilderTest.kt
DmitryNekrasov: Agreed, but all five builder iterators throw it bare, and none of them is touched by this PR. The message also differs by method: `remove()` needs a preceding `next()`, while `set()` on a list iterator is valid after `previous()` too. Filed as #352 to do all five at once.

DmitryNekrasov:
I moved #320, #329, #334 and #336 here as separate commits, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #328, opened 2026-08-21
retainingAll replaces the stored element instance when the receiver holds the element in a deeper node

### LINKED ISSUE #322: retainingAll replaces the stored element instance when the receiver holds the element in a deeper node

DmitryNekrasov:
The collision guard mirrors #321's. However, where the receiver holds a node and the argument holds the element, the fix is a lookup, not a lift-and-merge, since the intersection of a subtree with a single element is at most that element.

Fixes #322

### DIFF
--- a/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
+++ b/core/commonMain/src/implementations/immutableMap/PersistentHashMapBuilderContentIterators.kt
@@ -53,6 +53,7 @@ internal open class PersistentHashMapBuilderBaseIterator<K, V, T>(
 
     override fun remove() {
         checkNextWasInvoked()
+        checkForComodification()
         if (hasNext()) {
             val currentKey = currentKey()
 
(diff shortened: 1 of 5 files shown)


DmitryNekrasov:
Moved to #317 as a separate commit, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #329, opened 2026-08-22
remove on a hash set builder iterator throws an AssertionError after a remove of a different element

### LINKED ISSUE #325: remove on a hash set builder iterator throws an AssertionError after a remove of a different element

DmitryNekrasov:
`remove()` was the only mutator on the set iterator without the comodification check `next()` carries. `assert` is a no-op on JS and Wasm, so there the reported `AssertionError` does not appear at all - `remove()` empties the builder and the following `next()` returns `undefined`.

Fixes #325


ilya-g (dismissed):

DmitryNekrasov:

PR #334, opened 2026-08-23
remove on an ordered map builder iterator succeeds after a remove of a different key

### LINKED ISSUE #326: remove on an ordered map builder iterator succeeds after a remove of a different key

DmitryNekrasov:
`remove()` was the only method on the ordered map builder's links iterator without the comodification check `next()` carries. Unlike the two twins it never trips an assert - the builder splices both neighbours on every removal, so the chain stays consistent and the entry is just silently dropped.

Fixes #326

### DIFF
--- a/core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
+++ b/core/commonMain/src/implementations/persistentOrderedMap/PersistentOrderedMapBuilderContentIterators.kt
@@ -39,6 +39,7 @@ internal open class PersistentOrderedMapBuilderLinksIterator<K, V>(
 
     override fun remove() {
         checkNextWasInvoked()
+        checkForComodification()
         builder.remove(lastIteratedKey)
         lastIteratedKey = null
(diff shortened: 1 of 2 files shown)


DmitryNekrasov:

PR #336, opened 2026-08-24
remove on an ordered set builder iterator succeeds after a remove of a different element

### LINKED ISSUE #327: remove on an ordered set builder iterator succeeds after a remove of a different element

DmitryNekrasov:
`remove()` was the only method on the ordered set builder's iterator without the comodification check `next()` carries.

Fixes #327


DmitryNekrasov:

PR #337, opened 2026-08-24
hasNext on the ordered and list builder iterators ends the iteration early after a remove or an add through the builder

### LINKED ISSUE #332: hasNext on an ordered map builder iterator ends the iteration early after a remove of a visited key
A `while (iterator.hasNext())` loop over `persistentMapOf(...).builder()` ends silently after some key is removed directly from the builder, where `LinkedHashMap` and the hash map builder both throw `ConcurrentModificationException`. The entries after the cursor are never returned.

To reproduce:

```kotlin
iterator.next()
iterator.next()

builder.remove(1)

while (iterator.hasNext()) {  // false at once
    println(iterator.next())  // never runs, the entry for 3 is lost
}
```

`hasNext()` is wrong in the other direction too. On an exhausted iterator, after a new key is put into the builder, it answers true and the loop then throws `ConcurrentModificationException`, where `LinkedHashMap` finishes cleanly:

```kotlin
repeat(3) { iterator.next() }

builder[9] = "z"

while (iterator.hasNext()) {  // true
    println(iterator.next())  // ConcurrentModificationException
}
```

Found while fixing #326.

### LINKED ISSUE #335: hasNext on an ordered set builder iterator ends the iteration early after a remove of a visited element
A `while (iterator.hasNext())` loop over `persistentSetOf(...).builder()` ends silently after some element is removed directly from the builder, where `LinkedHashSet` and the hash set builder both throw `ConcurrentModificationException`. The elements after the cursor are never returned.

To reproduce:

```kotlin
iterator.next()
iterator.next()

builder.remove(1)

while (iterator.hasNext()) {  // false at once
    println(iterator.next())  // never runs, 3 is lost
}
```

`hasNext()` is wrong in the other direction too. On an exhausted iterator, after a new element is added to the builder, it answers true and the loop then throws `ConcurrentModificationException`, where `LinkedHashSet` and the hash set builder both finish cleanly:

```kotlin
repeat(3) { iterator.next() }

builder.add(9)

while (iterator.hasNext()) {  // true
    println(iterator.next())  // ConcurrentModificationException
}
```

The ordered map twin is #332.

Found while fixing #327.

### LINKED ISSUE #344: hasNext on a persistent list builder iterator ends the iteration early after an add through the builder
A `while (iterator.hasNext())` loop over `persistentListOf(...).builder()` ends silently after a new element is added directly to the builder, where `ArrayList` answers true and the loop then throws `ConcurrentModificationException`. The added element is never returned.

To reproduce:

```kotlin
repeat(3) { iterator.next() }

builder.add(9)

while (iterator.hasNext()) {  // false
    println(iterator.next())  // never runs, 9 is lost
}
```

`hasNext()` is wrong in the other direction too. After the last element is removed directly from the builder, it answers true and the loop then throws `ConcurrentModificationException`, where `ArrayList` finishes cleanly:

```kotlin
iterator.next()
iterator.next()

builder.removeAt(2)

while (iterator.hasNext()) {  // true
    println(iterator.next())  // ConcurrentModificationException
}
```

This is the list side of #332 and #335.

Found while fixing #340.

DmitryNekrasov:
`hasNext()` on the iterators of both ordered builders compared a count of `next()` calls with the live size of the builder, where the hash builders' iterators answer from their own path. The read-only map iterator gets the same check, so that all 4 ordered iterators use one.

`hasNext()` of the list builder's iterator compared the cursor with the size taken at the last reset. Now it uses the live size, as in `ArrayList`.

Fixes #332
Fixes #335
Fixes #344


DmitryNekrasov:
I moved #346 here as a separate commit, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #339, opened 2026-08-24
put of an existing key on a hash map builder and set on a list builder invalidate a live iterator depending on how the builder was created

### LINKED ISSUE #333: remove on a hash map builder iterator throws after a value overwrite, depending on how the builder was created
`remove()` on an iterator of `persistentHashMapOf(...).builder()` throws `ConcurrentModificationException` after a value is overwritten in the builder, but only for some builders. `HashMap` never throws there, and neither does the ordered map builder.

To reproduce:

```kotlin
val builder = persistentHashMapOf(1 to "a").builder()
iterator.next()

builder[1] = "b"

iterator.remove()  // ConcurrentModificationException
```

The same code on a builder filled by puts succeeds, and so does a second overwrite of the same key:

```kotlin
val builder = persistentHashMapOf<Int, String>().builder()
builder[1] = "a"
iterator.next()

builder[1] = "b"

iterator.remove()  // succeeds
```

So the same sequence of calls gives two different results depending on how the builder was reached.

Found while fixing #326.

### LINKED ISSUE #340: set on a persistent list builder invalidates a live iterator depending on how the builder was created
`next()` on an iterator of `persistentListOf(...).builder()` throws `ConcurrentModificationException` after an element is replaced through the builder with `set`, but only for some builders. `ArrayList` never throws there.

To reproduce:

```kotlin
iterator.next()

builder[1] = 20

iterator.next()  // ConcurrentModificationException
```

The same code on a builder filled by `add` succeeds and returns `20`:

```kotlin
builder.addAll(listOf(1, 2, 3))
iterator.next()

builder[1] = 20

iterator.next()  // 20
```

The same on JVM, JS and Native. This is the list side of #333.

DmitryNekrasov:
The hash map builder's iterators checked `modCount`, which also moves when an overwrite copies a node the builder does not own. Now they check `sizeModCount`, as the ordered map builder's iterators do since #311, and after a copy they move their path to the current key. A `put` or `putAll` that only replaces values no longer throws during iteration, as in `HashMap`.

While fixing, I noticed 2 more problems. The iterator kept a copy of the entry that its own `remove` moved up into the parent node, so a later overwrite of it was not seen. Also, a `putAll` that only replaces values could reuse the argument's collision node with the keys in another order while an iterator was inside that node.

The list builder had the same problem with `set` and gets the same fix. On the JVM `subList` views read the same `modCount`, so they no longer throw after a `set` either.

Fixes #333
Fixes #340


DmitryNekrasov:
We discussed at the meeting whether an overwrite through the builder should throw CME or let the iterator continue, and I think that continuing and returning the new value is the best option. An overwrite is not a structural modification by any written definition. [`HashMap`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/java/util/HashMap.java#L92-L95) says directly that "merely changing the value associated with a key that an instance already contains is not a structural modification". The stdlib's own definition on [`AbstractMutableList.modCount`](https://github.com/JetBrains/kotlin/blob/v2.3.0/libraries/stdlib/common/src/kotlin/collections/AbstractMutableList.kt#L15-L18) is wider. It says a modification is also structural if "iterations in progress may return incorrect results" after it. Before this PR an overwrite that copied a node was such a modification, because the iterator kept the buffers of the old node and would return the old value. Now the iterator rebuilds its path on the current trie after a copy and returns the current value, so this definition does not cover the overwrite either. The only java map that throws here, `LinkedHashMap` in access order, throws because the iteration order changes. An overwrite does not change the iteration order of a hash trie. Almost all java maps continue and return the new value, and the stdlib's own `ConcurrentModificationTest` checks that `put(update value)` does not throw on all platforms.

DmitryNekrasov:
I moved #343 here as a separate commit, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #343, opened 2026-08-28
set on a persistent list builder invalidates a live iterator depending on how the builder was created

### LINKED ISSUE #340: set on a persistent list builder invalidates a live iterator depending on how the builder was created

DmitryNekrasov:
`set` bumped `modCount` only when it had to copy a buffer the builder does not own, so whether a live iterator survived a `set` depended on where the builder came from. `modCount` now counts only size changes, as in the hash map builder since #339, and the iterator rebuilds its cached trie path after a `set` that copied a leaf, so `next()` and `previous()` return the new element in both flavours. On the JVM `subList` views read the same field and follow a `set` now too.

Fixes #340

### DIFF
--- a/core/commonMain/src/implementations/immutableList/PersistentVectorBuilder.kt
+++ b/core/commonMain/src/implementations/immutableList/PersistentVectorBuilder.kt
@@ -42,6 +42,10 @@ internal class PersistentVectorBuilder<E>(
 
     internal fun getModCount() = modCount
 
+    // Counts the sets that replaced a trie buffer with its copy; the iterators re-anchor their cached path on it.
+    internal var trieModCount = 0
+        private set
+
     override fun build(): PersistentList<E> {
         return builtVector ?: run {
             val root = root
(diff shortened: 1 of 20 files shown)


DmitryNekrasov:
Moved to #339 as a separate commit, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #346, opened 2026-09-01
hasNext on a persistent list builder iterator ends the iteration early after an add through the builder

### LINKED ISSUE #344: hasNext on a persistent list builder iterator ends the iteration early after an add through the builder

DmitryNekrasov:
hasNext of the builder's iterator now answers from the live size, as in `ArrayList`. A cursor left beyond the live size still reports an element - the loop does not end silently in this state either, and `next()` reports the comodification the same way `previous()` already does.

Fixes #344

### DIFF
--- a/core/commonMain/src/implementations/immutableList/PersistentVectorMutableIterator.kt
+++ b/core/commonMain/src/implementations/immutableList/PersistentVectorMutableIterator.kt
@@ -46,6 +46,13 @@ internal class PersistentVectorMutableIterator<T>(
         setupTrieIterator()
     }
 
+    /**
+     * Answers true for a cursor left beyond the live size, leaving the comodification for [next] to report.
+     */
+    override fun hasNext(): Boolean {
+        return index != builder.size
+    }
+
     override fun previous(): T {
         checkForComodification()
         checkHasPrevious()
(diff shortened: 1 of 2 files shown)


DmitryNekrasov:
Moved to #337 as a separate commit, see https://github.com/Kotlin/kotlinx.collections.immutable/pull/317#issuecomment-5542373533.

PR #347, opened 2026-09-02
removeAll on a persistent list builder duplicates an element when contains throws

### LINKED ISSUE #345: removeAll on a persistent list builder duplicates an element when contains throws
`removeAll` on a builder filled by `add` leaves it corrupted when the argument's `contains` throws partway: a surviving element is duplicated and the size does not change. `ArrayList` in the same call removes the elements checked before the throw and keeps the rest intact.

To reproduce:

```kotlin
builder.addAll(listOf(1, 2, 3, 4))
val elements = object : AbstractCollection<Int>() {
    override val size = 1
    override fun iterator() = listOf(1).iterator()
    override fun contains(element: Int): Boolean {
        if (element == 3) throw IllegalStateException()
        return element == 1
    }
}

try {
    builder.removeAll(elements)
} catch (e: IllegalStateException) {
}

println(builder)  // [2, 2, 3, 4], ArrayList gives [2, 3, 4]
```

The same call on a builder taken from a persistent list leaves it unchanged.

Found while fixing #340.

DmitryNekrasov:
`removeAll` compacts owned buffers in place while it still calls `contains`, so a throw left the buffer half compacted. The fix is to rethrow only after the pass has finished, so the elements already reported as contained are removed and the rest stay in place, as in `java.util.ArrayList` and as `retainAll` on the same builder already does.

Fixes #345


core/commonTest/src/contract/list/PersistentListBuilderTest.kt
```diff
@@ -199,4 +199,82 @@ class PersistentListBuilderTest {
             assertFailsWith<ConcurrentModificationException>(flavour) { iterator.next() }
         }
     }
+
+    @Test
+    fun `removeAll whose contains throws removes the elements matched before the throw and keeps the rest at every depth`() {
```
fzhinkin, line 204: I don't get what it says, sorry. What is removeAll's "contains"? A predicate?

fzhinkin: Ah, I got it, but the test's name is unclear and somewhat misleading.

core/commonMain/src/implementations/immutableList/PersistentVectorBuilder.kt
```diff
@@ -720,10 +720,19 @@ internal class PersistentVectorBuilder<E>(
 
     @IgnorableReturnValue
     fun removeAllWithPredicate(predicate: (E) -> Boolean): Boolean {
```
fzhinkin, line 722: What are the guarantees about builder's behavior and exceptions thrown from the collection?

What about other operations? Like `addAll` accepted a faulty collection.

PR #350, opened 2026-09-04
remove and set on a list builder iterator throw ConcurrentModificationException when next was never called

### LINKED ISSUE #331: remove and set on a list builder iterator throw ConcurrentModificationException when next was never called
`remove()` and `set()` on an iterator of a list builder report a comodification even when `next()` was never called. The hash map and hash set builder iterators throw `IllegalStateException` here, as do [`MutableListIterator.remove`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-mutable-list-iterator/remove.html) and [`set`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-mutable-list-iterator/set.html) in their documentation.

To reproduce:

```kotlin
val builder = persistentListOf(1, 2).builder()
val iterator = builder.listIterator()

builder.add(3)

iterator.remove()  // ConcurrentModificationException, expected IllegalStateException
```

`iterator.set(9)` in place of `remove()` behaves the same.

The stdlib does the same on Native and Wasm, which is [KT-88775](https://youtrack.jetbrains.com/issue/KT-88775).

DmitryNekrasov:
`PersistentVectorMutableIterator` was the only builder iterator checking for a comodification before the last returned element - the hash map, hash set, ordered map and ordered set iterators all check it the other way round, as `java.util.ArrayList` does.

Fixes #331


core/commonTest/src/contract/list/PersistentListBuilderTest.kt
```diff
@@ -200,6 +200,162 @@ class PersistentListBuilderTest {
...
+            builder.add(3)
+
+            assertFailsWith<IllegalStateException>(flavour) { iterator.remove() }
+            assertFailsWith<IllegalStateException>(flavour) { iterator.set(-1) }
+            assertFailsWith<ConcurrentModificationException>(flavour) { iterator.next() }
+        }
+    }
+
+    @Test
+    fun `iterator remove and set on an untouched builder throw IllegalStateException`() {
```
fzhinkin, line 230: >  the hash map, hash set, ordered map and ordered set iterators all check it the other way round

I guess, the same behavior is expected from other builders too?

That imply that:
1) There should be tests verifying this behavior for other collections too. However, I picked this test and checked `PersistentOrderedMapBuilderTest` - there's no corresponding test.
2) It's a contract/behavior common to all collections and builders providing mutable iterator into them. If so, it makes sense to implement a single test for that common contract and parameterize it with different iterable implementations instead of reimplementing the same test multiple times. You can check Kotlin stdlib's tests covering collections behavior for the inspiration.

core/commonTest/src/contract/list/PersistentListBuilderTest.kt
```diff
@@ -200,6 +200,162 @@ class PersistentListBuilderTest {
...
+            builder.removeAt(0)
+
+            assertFailsWith<ConcurrentModificationException>(flavour) { iterator.remove() }
+            assertFailsWith<ConcurrentModificationException>(flavour) { iterator.set(-1) }
+            assertEquals(listOf(2), builder.toList(), flavour)
+        }
+    }
+
+    @Test
+    fun `next and previous throw ConcurrentModificationException rather than NoSuchElementException when the iteration cannot continue`() {
```
fzhinkin, line 340: While there are certain cases where long elaborative case names are necessary, this is not one of them. The test case name should convey what test is all about, but is should not be a literal description of code inside the test body.

Here, `rather than NoSuchElementException` is clearly unnecessary. It could list all known exceptions, or none, that wouldn't affect the semantics - the test only checks for `ConcurrentModificationException`.
Then, there's clearly two separate cases - one for `next`, another one for `previous`.

Further scoping of test classes can take a significant part of the name on it.

For example:
```
class PersistentListBuilderTest {
   class IterationTests {
      class ConcurrentModificationTests {
           @Test fun `next at list tail`() { ... }
           @Test fun `previous at list head`() { ... }
      }
   }
}
```

core/commonTest/src/contract/list/PersistentListBuilderTest.kt
```diff
@@ -200,6 +200,162 @@ class PersistentListBuilderTest {
...
+            builder.add(3)
+
+            assertFailsWith<IllegalStateException>(flavour) { iterator.remove() }
+            assertEquals(listOf(1, 2, 3), builder.toList(), flavour)
+        }
+
+        for ((flavour, builder) in builders(3)) {
+            val iterator = builder.listIterator()
+            assertEquals(0, iterator.next(), flavour)
+            iterator.add(-1)
```
fzhinkin, line 283: There are generally no positive tests for mutable iterator's operations, besides a few added here. I understand that these cases are regression tests of some sort, but what's the general plan here?

core/commonTest/src/contract/list/PersistentListBuilderTest.kt
```diff
@@ -200,6 +200,162 @@ class PersistentListBuilderTest {
...
+            builder.add(3)
+
+            assertFailsWith<IllegalStateException>(flavour) { iterator.remove() }
+            assertFailsWith<IllegalStateException>(flavour) { iterator.set(-1) }
+            assertEquals(listOf(0, 1, 2, 3), builder.toList(), flavour)
+        }
+    }
+
+    @Test
+    fun `next after a remove and a set that threw IllegalStateException still throws ConcurrentModificationException`() {
```
fzhinkin, line 217: Looking at tests, it is clear that it would be much easier to reason about the behavior if any operation on the iterator invalidated by collection's modification would throw the CME. But if there are strong reasons to have this particular behavior, so be it.

core/commonTest/src/contract/list/PersistentListBuilderTest.kt
```diff
@@ -200,6 +200,162 @@ class PersistentListBuilderTest {
...
+    }
+
+    @Test
+    fun `iterator remove and set on an untouched builder throw IllegalStateException`() {
+        for ((flavour, builder) in builders(3)) {
+            val iterator = builder.listIterator()
+
+            assertFailsWith<IllegalStateException>(flavour) { iterator.remove() }
+            assertFailsWith<IllegalStateException>(flavour) { iterator.set(-1) }
+            assertEquals(listOf(0, 1, 2), builder.toList(), flavour)
```
fzhinkin, line 236: It seems like this case and `iterator remove and set without a preceding next after an external add throw IllegalStateException` differs only by the number of operations applied to a builder. The same is true for other tests where only the position of iterator differs from case to case.

Perhaps, instead of copy-pasting (or generating) a set of almost identical tests, it's better to write a couple that will iteratively modify list, advance/retreat iterator, and check the behavior.
Note that it does not mean that the checks has to be performed on the exactly the same instance of the builder/iterator, but there could be a set of operations and the test frameworks would do everything else.

fzhinkin (changes requested):

PR #359, opened 2026-09-09
Raise the imperative names of persistent collections to DeprecationLevel.ERROR

DmitryNekrasov:
Following the [migration guide](https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/docs/0.5.0-MIGRATION.md), I raised the deprecation level of the old methods to `ERROR` and moved the implementation into the participial forms.


core/commonMain/src/extensions.kt
```diff
@@ -316,7 +316,8 @@ public fun <K, V> PersistentMap<out K, V>.puttingAll(map: Map<out K, V>): Persis
  */
 @Deprecated(
     "Use puttingAll() instead. For more details, read the documentation for this function.",
-    replaceWith = ReplaceWith("puttingAll(map)")
+    replaceWith = ReplaceWith("puttingAll(map)"),
+    level = DeprecationLevel.ERROR
```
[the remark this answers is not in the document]
DmitryNekrasov: Done.

docs/0.5.0-MIGRATION.md
```diff
@@ -83,7 +83,7 @@ upgrade.
 | Version | `DeprecationLevel` | What happens | Action required |
 | --- | --- | --- | --- |
 | `0.5.0` | `WARNING` — introduced in [#233](https://github.com/Kotlin/kotlinx.collections.immutable/pull/233) | Old names still compile. IDE shows deprecation warnings with `ReplaceWith` quick-fixes. Binary compatible. | None immediately. Start migrating at your own pace. |
-| `0.6.0` | `ERROR` | Old names still resolve at runtime (binary compatible) but no longer compile from source. | Migrate all call sites before upgrading. |
+| `0.6.0` | `ERROR` | Old names still resolve at runtime, so already compiled callers keep working, but they no longer compile from source. Implementations of the interfaces must override the participial methods. | Migrate all call sites and implementations before upgrading. |
```
fzhinkin, line 86:
```suggestion
| `0.6.0` | `ERROR` | Old names still resolve at runtime, so already compiled callers keep working, but they no longer compile from source. Implementations of the interfaces must override the participial functions. | Migrate all call sites and implementations before upgrading. |
```

It's functions, not methods.

> the participial methods

Also, it might be better to refer them as "new" / "recently added" functions, or "non-deprecated" functions. It's much easier to get without looking other docs.

fzhinkin: Otherwise, the change looks good

DmitryNekrasov: Done. Replaced method with function across the guide. The functions are now called new and old, participial stays only where the naming rule itself is explained.

fzhinkin (approved):

docs/0.5.0-MIGRATION.md
fzhinkin: Thanks!

